PR #1, opened 2025-08-27
Initial MCP server setup

MacondoExpress:
# Initial Neo4j MCP Project Setup

## Summary
This PR establishes the foundational structure for the Neo4j Model Context Protocol (MCP) server implementation. It includes a basic prototype of the `read-cypher` tool to demonstrate core functionality, but is not production-ready.

**Note**: This PR also serves to test the repository workflow and CI/CD processes on a non-empty project structure.

This is a **placeholder implementation** and lacks several production requirements.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -9,8 +9,8 @@ import (
 
 	"github.com/mark3labs/mcp-go/mcp"
 	"github.com/mark3labs/mcp-go/server"
-	"github.com/neo4j/neo4j-go-driver/v5/neo4j"
 	"github.com/neo4j/mcp/internal/config"
+	"github.com/neo4j/neo4j-go-driver/v5/neo4j"
 )
 
 // Neo4jMCPServer represents the MCP server instance
@@ -21,13 +21,11 @@ type Neo4jMCPServer struct {
 }
 
 // NewNeo4jMCPServer creates a new MCP server instance
-func NewNeo4jMCPServer() *Neo4jMCPServer {
+func NewNeo4jMCPServer(cfg *config.Config) *Neo4jMCPServer {
 	mcpServer := server.NewMCPServer(
 		"neo4-mcp",
 		"0.0.1",
 	)
-
-	cfg := config.LoadConfig()
 
 	// Initialize Neo4j driver once
 	driver, err := neo4j.NewDriverWithContext(cfg.URI, neo4j.BasicAuth(cfg.Username, cfg.Password, ""))
@@ -98,6 +96,7 @@ func (s *Neo4jMCPServer) handleReadCypher(ctx context.Context, request mcp.CallT
 	// Execute the Cypher query using the stored driver
 	response, err := s.executeReadQuery(ctx, Query, Params)
 	if err != nil {
+		// TODO: discuss and write guideline on how to handle tool calling errors.
 		return mcp.NewToolResultError(fmt.Sprintf("Error executing Cypher query: %v", err)), nil
 	}
 
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,6 @@
 module github.com/neo4j/mcp
 
-go 1.24.0
-
-toolchain go1.24.4
+go 1.25.0
 
 require (
 	github.com/mark3labs/mcp-go v0.38.0
--- a/main.go
+++ b/main.go
@@ -1,12 +1,16 @@
 package main
 
 import (
+	"github.com/neo4j/mcp/internal/config"
 	"github.com/neo4j/mcp/internal/server"
 )
 
 func main() {
+	// get config from environment variables
+	cfg := config.LoadConfig()
+
 	// Create and configure the MCP server
-	mcpServer := server.NewNeo4jMCPServer()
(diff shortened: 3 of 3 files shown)


internal/server/server.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// NewNeo4jMCPServer creates a new MCP server instance
+func NewNeo4jMCPServer() *Neo4jMCPServer {
+	mcpServer := server.NewMCPServer(
+		"neo4-mcp",
+		"0.0.1",
+	)
+
+	cfg := config.LoadConfig()
```
angrykoala: Consider passing the cfg as a parameter rather than loading it here

internal/server/server.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// Define ReadCypher input schema
+type ReadCypherInput struct {
+	Query  string         `json:"query" jsonschema:"default=MATCH(n) RETURN n,description=The Cypher query to execute"`
+	Params map[string]any `json:"params" jsonschema:"description=Parameters to pass to the Cypher query"`
+}
+
+// RegisterTools registers all available MCP tools
+func (s *Neo4jMCPServer) RegisterTools() {
```
angrykoala, line 50: This seems to be all dealing with private information, any reason for not doing this in the Start function?

internal/server/server.go
```diff
@@ -0,0 +1,136 @@
...
+	if Params != nil {
+		fmt.Fprintf(os.Stderr, "cypher-parameters: %v\n", Params)
+	}
+	fmt.Fprintf(os.Stderr, "url: %s\n", s.config.URI)
+	fmt.Fprintf(os.Stderr, "database: %s\n", s.config.Database)
+
+	// Execute the Cypher query using the stored driver
+	response, err := s.executeReadQuery(ctx, Query, Params)
+	if err != nil {
+		return mcp.NewToolResultError(fmt.Sprintf("Error executing Cypher query: %v", err)), nil
```
keremgocen, line 100: I think this line should return the err instead of wrapping it inside the NewToolResultError? or was that intentional?

internal/server/server.go
```diff
@@ -0,0 +1,136 @@
...
+	ctx := context.Background()
+	return (*s.driver).Close(ctx)
+}
+
+// handleReadCypher handles the read-cypher tool requests
+func (s *Neo4jMCPServer) handleReadCypher(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+	var args ReadCypherInput
+	// Bind arguments to the struct
+	if err := request.BindArguments(&args); err != nil {
+		return mcp.NewToolResultError(err.Error()), nil
```
keremgocen, line 84: same here about err return being in reverse order, it feels more natural to return the error and bubble it up to be handled at the upper level in golang, I might need to check how github.com/mark3labs/mcp-go/mcp handles with errors as well.

internal/server/server.go
```diff
@@ -0,0 +1,136 @@
...
+	var results []map[string]any
+	for _, record := range res.Records {
+		recordMap := make(map[string]any)
+		for i, key := range record.Keys {
+			recordMap[key] = record.Values[i]
+		}
+		results = append(results, recordMap)
+	}
+
+	formattedResponse, err := json.MarshalIndent(results, "", "  ")
```
keremgocen, line 125: the formatting here feels a bit brittle, I'd move it to a separate function and add some tests against to make sure it behaves as we expect and modify it easily but the whole query execution logic is in progress I'm guessing right now. (then this function also becomes `excute` -> `parse` -> `handle error if any or return`)

internal/server/server.go
MacondoExpress: It will soon be changed to handle more logic, registering a different set of tools depending on the configuration/environment. Also, likely in the future, there will also be other register functions such as "registerPrompts", "registerResources", etc Depending on the future implementation, I will expose these "register" functions or just invoke them from the "Start" method.

internal/server/server.go
MacondoExpress: It was intentional, 
I wrapped any errors happening in the tool inside the "NewToolResultError" from the mark3labs method signature:
```
// NewToolResultError creates a new CallToolResult with an error message.
// Any errors that originate from the tool SHOULD be reported inside the result object.
```
From the spec:
https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling
I'm not 100% sure here, I feel is the "most right" given the nature of the tool, however, I think we should have a conversation/guidelines regarding the error handling, especially for benefit to external contributors! For instance, maybe in this case Cypher errors are handled correctly, but "network error" not. 
Thanks for raising it!

internal/server/server.go
MacondoExpress: same as above.

internal/server/server.go
MacondoExpress: Definitely,  it's already in the plan, I'll keep it out of the scope of this PR as this PR was mainly to have a starting point to test the repository ergonomics.

PR #2, opened 2025-08-28
Add get-schema and continue project initialization

MacondoExpress:
# Restructure project and add get-schema tool

This PR is a follow-up of the initial project initialization and includes:

- Moved `main.go` to `/cmd/neo4-mcp`
- Refactored tools into separate packages with ergonomic register_tools to allow to write and register tools using "spec" file and "handler" file
- Added run-cypher tool for write operations (replaces read-only read-cypher)
- Added get-schema tool using APOC for database schema retrieval
- Updated README with comprehensive setup and installation instructions
- Extract shared functionalities in separate folders/files.

### DIFF
--- a/internal/database/execute_query.go
+++ b/internal/database/execute_query.go
@@ -22,7 +22,7 @@ func ExecuteReadQuery(ctx context.Context, driver *neo4j.DriverWithContext, cyph
 
 // ExecuteWriteQuery executes a write-only Cypher query and returns raw records
 func ExecuteWriteQuery(ctx context.Context, driver *neo4j.DriverWithContext, cypher string, params map[string]any, database string) ([]*neo4j.Record, error) {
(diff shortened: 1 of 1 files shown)


internal/database/execute_query.go
```diff
@@ -0,0 +1,33 @@
...
+		fmt.Fprintf(os.Stderr, "Error while executing Cypher: %v\n", err)
+		return nil, err
+	}
+
+	return res.Records, nil
+}
+
+// ExecuteWriteQuery executes a write-only Cypher query and returns raw records
+func ExecuteWriteQuery(ctx context.Context, driver *neo4j.DriverWithContext, cypher string, params map[string]any, database string) ([]*neo4j.Record, error) {
+	res, err := neo4j.ExecuteQuery(ctx, *driver, cypher, params, neo4j.EagerResultTransformer, neo4j.ExecuteQueryWithDatabase(database))
```
Liam-Doodson: Would it be better to explicitly define the write routing?

MacondoExpress: Fair enough!

PR #4, opened 2025-09-04
This introduces a change to config.go to support the use of .env file…

LackOfMorals:
… that holds the needed environmental variables and their values

A new function, getEnvFilePath, has been written that returns the full folder path to where the mcp binary is located as this is where we expect the .env file to be.

getEnvFilePath uses filepath.Abs(os.Args[0]) which to obtain full folder path and the binary filename which is then split with filepath.Split to obtain the full folder path which is then returned.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -16,34 +16,59 @@ type Config struct {
 	Database string
 }
 
-func getEnvFilePath() string {
+// Returns the full filepath to the neo4j-mcp binary or an error on failure.
+func getEnvFilePath() (string, error) {
 	// The .env file has to be in the same folder as the neo4j-mcp binary
 	// Get the filepath to the neo4j-mcp binary
-	binaryFullPath, _ := filepath.Abs(os.Args[0])
+
+	var err error
+	var binaryFullPath, envFolderPath string
+
+	binaryFullPath, err = filepath.Abs(os.Args[0])
+	if err != nil {
+		return "", err
+	}
 
 	// filepath.Abs returns full folder path and the binary filename but we only want the full folder path so we
 	// use filepath.Split that returns two values, dir = full folder path  and
 	// file = the binary name
 	// We just want the path which is where the .env file ( hopefully ) will be
-	envFolderPath, _ := filepath.Split(binaryFullPath)
+	envFolderPath, _ = filepath.Split(binaryFullPath)
 
-	return envFolderPath
+	return envFolderPath, nil
+}
+
+// Sets environment variables from .env file located in the same folder as the neo4j-mcp binary
+func setEnvFromFile() error {
+	var err error
+	var envFilePath, envFileName string
+
+	envFileName = ".env"
+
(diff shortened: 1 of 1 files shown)


internal/config/config.go
```diff
@@ -12,8 +16,36 @@ type Config struct {
 	Database string
 }
 
+func getEnvFilePath() string {
+	// The .env file has to be in the same folder as the neo4j-mcp binary
+	// Get the filepath to the neo4j-mcp binary
+	binaryFullPath, _ := filepath.Abs(os.Args[0])
```
keremgocen: ignoring errors from `filepath.Abs()` could be problematic, instead we should handle potential errors gracefully and change the function signature to something like `func getEnvFilePath() (string, error)` so we can bubble up errors to the caller, and return an empty string.

internal/config/config.go
```diff
@@ -12,8 +16,36 @@ type Config struct {
...
+
 // LoadConfig loads configuration from environment variables with defaults
 func LoadConfig() *Config {
+
+	// try loading .env and set the needed
+	// environmental variables
+	envFilePath := getEnvFilePath()
+	envFileName := ".env"
+
+	err := godotenv.Load(envFilePath + envFileName)
```
keremgocen: direct string concatenation may not be cross-platform safe maybe use `filepath.Join(envFilePath, envFileName)` instead
(see: https://pkg.go.dev/path/filepath#Join)

internal/config/config.go
```diff
@@ -12,8 +16,36 @@ type Config struct {
 	Database string
 }
 
+func getEnvFilePath() string {
```
keremgocen: do we want to add a description to this function?
(and maybe some basic test coverage since it's doing things with the path and we want to run it as part of the CI later to test it against cross platform)

internal/config/config.go
```diff
@@ -12,8 +16,36 @@ type Config struct {
...
+	// try loading .env and set the needed
+	// environmental variables
+	envFilePath := getEnvFilePath()
+	envFileName := ".env"
+
+	err := godotenv.Load(envFilePath + envFileName)
+
+	// We didn't load the .env file :(
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Could  not find .env file in %s\n", envFilePath)
```
keremgocen:
```suggestion
		fmt.Fprintf(os.Stderr, "Could not find .env file in %s\n", envFilePath)
```
sorry I also can't ignore inconsistent spacing:)

internal/config/config.go
LackOfMorals: function now returns errors along with a description

internal/config/config.go
LackOfMorals: Changed to use filepath.join

internal/config/config.go
LackOfMorals: Description is now added.

PR #5, opened 2025-09-09
Add database service interfaces and mockgen with tests

keremgocen:
Introduce mock generation for database interfaces and implement a concrete database service. This enhances testing capabilities and provides a structured way to interact with Neo4j queries.

Some documentation updates (maybe too much).

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Create or update your VSCode MCP configuration file (`mcp.json`), as document he
   "servers": {
     "neo4j": {
       "type": "stdio",
-      "command": "./bin/neo4j-mcp", // Use full path to binary or ensure neo4j-mcp is in PATH
+      "command": "neo4j-mcp", // Use full path to binary or ensure neo4j-mcp is in PATH
       "env": {
         "NEO4J_URI": "bolt://localhost:7687",
         "NEO4J_USERNAME": "neo4j",


internal/server/server.go
```diff
@@ -20,49 +20,57 @@ type Neo4jMCPServer struct {
...
 }
 
 // RegisterTools registers all available MCP tools
-func (s *Neo4jMCPServer) RegisterTools() {
+func (s *Neo4jMCPServer) RegisterTools() error {
+	// Create the database service
+	dbService := database.NewNeo4jService(s.driver)
+
 	deps := &tools.ToolDependencies{
-		Driver: s.driver,
```
keremgocen, line 45: I'm not sure if we need the driver as a tool dependency or if it gives any additional context to the MCP server at all?

it runs fine without it apparently, it's more like a dependency of the database service connection imo.

LackOfMorals: I can't imagine what additional context having the driver as a tool dependency would give to the MCP Server.  

If we're unsure, and everything still works without it, then I'd not bother.

docs/database-interfaces.md
```diff
@@ -0,0 +1,76 @@
+# Database Interface Testing
+
+This guide covers the database interfaces and their generated mocks for testing.
+
+## Interface Structure
```
keremgocen: I think we don't really benefit from too many docs at this point and it might be confusing. I merely wanted to keep this for knowledge sharing so I'd like to merge these into a single readme I guess.

internal/database/interfaces.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"context"
+
+	"github.com/neo4j/neo4j-go-driver/v5/neo4j"
+)
+
+// QueryExecutor defines the interface for executing Neo4j queries
+type QueryExecutor interface {
+	// ExecuteReadQuery executes a read-only Cypher query and returns raw records
+	ExecuteReadQuery(ctx context.Context, cypher string, params map[string]any, database string) ([]*neo4j.Record, error)
```
keremgocen, line 14: this is what I wanted to have before adding more tests, defining the behaviour in interfaces like these will help us maintain it easier and generate mocks for interface-based dependency injection in tests etc. (ie. we probably don't want to test neo4j database or the driver but how the MCP server behaves)

so the goal is the server handles MCP protocol and lifecycle management while database operations are delegated to service layer through interfaces.

cmd/neo4j-mcp/main.go
```diff
@@ -10,12 +13,21 @@ func main() {
 	cfg := config.LoadConfig()
 
 	// Create and configure the MCP server
-	mcpServer := server.NewNeo4jMCPServer(cfg)
-	defer mcpServer.Stop()
-
-	// Register all tools
-	mcpServer.RegisterTools()
+	mcpServer, err := server.NewNeo4jMCPServer(cfg)
```
keremgocen, line 16: this will help propagate errors up the stack when debugging, I also added context to the mcpServer functions so we can use it for shutdown or timeouts etc.

internal/server/server.go
```diff
@@ -20,49 +20,57 @@ type Neo4jMCPServer struct {
...
-	fmt.Fprintf(os.Stderr, "Starting Cypher MCP Server running on stdio\n")
+// Start initializes and starts the MCP server using stdio transport
+func (s *Neo4jMCPServer) Start(ctx context.Context) error {
+	log.Println("Starting Neo4j MCP Server...")
 
-	// Start the server with stdio transport (this is blocking)
-	if err := server.ServeStdio(s.mcpServer); err != nil {
-		return fmt.Errorf("server error: %w", err)
+	// Test the database connection
+	if err := (*s.driver).VerifyConnectivity(ctx); err != nil {
```
keremgocen, line 63: tried to verify we have a connection to the db before starting the server but we might want to double check how this is used in other repos.

.gitignore
```diff
@@ -29,8 +29,13 @@ go.work.sum
...
+.vscode/
 
 
 # Binaries
 bin/
+
+# Generated mocks (auto-generated by mockgen)
+**/mocks/mock_*.go
+
+neo4j-mcp
```
mjfwebb: What's this? Can we get a comment for it please?

internal/server/server.go
```diff
@@ -20,49 +20,57 @@ type Neo4jMCPServer struct {
...
 	tools.RegisterAllTools(s.mcpServer, deps)
+	return nil
 }
 
-// Start starts the MCP server with stdio transport
-func (s *Neo4jMCPServer) Start() error {
-	fmt.Fprintf(os.Stderr, "Starting Cypher MCP Server running on stdio\n")
+// Start initializes and starts the MCP server using stdio transport
+func (s *Neo4jMCPServer) Start(ctx context.Context) error {
+	log.Println("Starting Neo4j MCP Server...")
```
mjfwebb: What's the reasoning behind changing how the logging is performed?

README.md
```diff
@@ -65,6 +65,10 @@ Adjust the environment variables according to your Neo4j instance configuration.
 
 Open the VSCode chat in agentic mode and ask about your configured Neo4j database.
 
+## Documentation
+
+📚 **[Development Guide](docs/README.md)** - Testing, mockgen setup, and development workflows
```
mjfwebb, line 70: Nit: maybe call the new file something other than README.md?

Liam-Doodson: In our other open source repos we have CONTRIBUTING.md at the root of the repo. Might be nice to keeps consistent with that

Liam-Doodson: Now I've read these docs maybe DEVELOPING.md or something like that could work?

docs/README.md
```diff
@@ -0,0 +1,60 @@
...
+See `internal/tools/get_schema_handler_gomock_test.go` for complete examples.
+
+## Project Structure
+
+```
+cmd/neo4j-mcp/         # Main application
+internal/
+├── database/          # Database interfaces & generated mocks
+├── server/            # MCP server setup
+└── tools/             # MCP tool handlers
```
Liam-Doodson, line 55: One for a future PR but I do think it's worth thinking if the current structure is well suited to multiple teams adding their tooling. Maybe we could say every team/sub server get's its section in tools?

internal/database/service.go
```diff
@@ -0,0 +1,74 @@
...
+func (s *Neo4jService) Neo4jRecordsToJSON(records []*neo4j.Record) (string, error) {
+	var results []map[string]any
+	for _, record := range records {
+		recordMap := record.AsMap()
+		results = append(results, recordMap)
+	}
+
+	formattedResponse, err := json.MarshalIndent(results, "", "  ")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error formatting response as JSON: %v\n", err)
```
Liam-Doodson: You've changed to log elsewhere? Is it worth making a follow-up change that makes our logging more consistent?

internal/server/server.go
Liam-Doodson: Yeah I think this is a good shout. Better to fail on startup with a useful message that at runtime

README.md
keremgocen: I think it's worth doing another sweep to the whole doc folder in another PR. Having a CONTRIBUTING.md at the top level makes sense to me, docs folder is usually for API docs which is derived from the code but I've used it differently with swagger/openAPI docs as well in the past. For now I think my goal was to keep it simple and make sure to explain what/how we use mockgen.

docs/README.md
keremgocen: That sounds reasonable, I'm trying to find and read up on other servers and how they are organised as well, I think this structure is subject to change over time as we add more transports etc. https://github.com/metoro-io/mcp-golang/tree/main/transport. We can discuss this further once we are updating the docs in general.

internal/database/service.go
keremgocen: good catch, I think I meant to return a wrapped error instead of logging here (with https://pkg.go.dev/fmt#Errorf)

internal/server/server.go
keremgocen: log package automatically adds timestamps to output so that's easier to debug for server apps, however I also like to return errors with fmt.Errorf as it allows you to wrap them with a useful message before returning. But for logs it's better to use the standard `log` or the newer `slog` (which adds structured logging info,debug, etc.) packages imo.

.gitignore
keremgocen: after I built the server locally, it created a binary called `neo4j-mcp` at the root level for me and I wanted to avoid committing it to github. I'll add a comment:)

internal/server/server.go
mjfwebb: Sounds good, let's standardise it whichever way we go (possibly in a followup PR)

README.md
mjfwebb: Sounds good

internal/database/service.go
keremgocen: so yeah, now we are using the log package for debug logging and we are returning wrapped errors via `fmt.Errorf` to add more context to the error bubbling it up the stack without losing the original err, which makes the stack trace cleaner and easier to debug imo.

.gitignore
keremgocen: so I think I created this binary via `go build ./cmd/neo4j-mcp` when I was trying to get it running. If you follow the docs and use the `go build -C cmd/neo4j-mcp -o ../../bin/` to build the binary, it goes into the /bin folder which is already ignored here. I'll remove my change for now as it might be confusing and we can add it later if it becomes a problem.

internal/server/server.go
LackOfMorals: FYI 

Most MCP Clients used by LLMs / Agents use stderr to receive error type information from the MCP Server.     That's why fmt.Fprintf(os.Stderr, "Something happened ") has been used here.  

With Claude, this is also surfaced in the UI to the end user.

Liam-Doodson: This is a good point @LackOfMorals.

From the spec:

> The server MUST NOT write anything to its stdout that is not a valid MCP message.

Liam-Doodson: > The server MAY write UTF-8 strings to its standard error (stderr) for logging purposes. Clients MAY capture, forward, or ignore this logging.

keremgocen: sorry I noticed this very late, but Go's `log` package is by default outputs to STDERR, for some reason I was assuming it goes into STDOUT.... 🤦 

anyway, you can double check this via a quick test here https://go.dev/play/p/Xjj1jl3pk3R, I also tried adding some debug `log.Println()` logs and tested the output via `./bin/neo4j-mcp 2>stderr.log 1>stdout.log` (you can get another terminal up and do `tail -f stderr.log` or `tail -f stdout.log` to see what goes into where.

We actually don't output anything into STDOUT, and everything goes into STDERR.

Another advantage of using a centralised logging mechanism is, in a later PR, we can inject the logger to the service as a dependency which would allow users to pick the output and log level they way. Apparently you can set the output of the logger via https://pkg.go.dev/log#SetOutput. So that's good.

If you all are ok with it, I'd like to keep `log.Println` for our log messages, and move on with it for now. We can put a trello card or write a feature request for more flexible logging later.

mjfwebb: Considering the spec states that we `MUST NOT` write to stdout, I would like this to be considered a MUST HAVE from planning perspective, but I have no issue with work on the logger being done on a follow-up PR :+1: 

EDIT: Just to clarify, I acknowledge and understand we're not logging to STDOUT :+1:

EDIT 2: But I would like a centralised logging mechanism so that we are not accidentally logging to STDOUT in the future

Liam-Doodson: > But I would like a centralised logging mechanism so

Yeah this is definitely part of the plan

.gitignore
```diff
@@ -29,8 +29,11 @@ go.work.sum
...
 # .idea/
-# .vscode/
+.vscode/
 
 
 # Binaries
 bin/
+
+# Generated mocks (auto-generated by mockgen)
+**/mocks/mock_*.go
```
Liam-Doodson: What's the thought process in ignoring these? Do we not want everyone to have the same mocks?

(I've not used mockgen before so trying to understand as much as anything)

internal/database/service.go
```diff
@@ -0,0 +1,82 @@
...
+	formattedResponse, err := json.MarshalIndent(results, "", "  ")
+	if err != nil {
+		wrappedErr := fmt.Errorf("failed to format records as JSON: %w", err)
+		log.Printf("Error in Neo4jRecordsToJSON: %v", wrappedErr)
+		return "", wrappedErr
+	}
+
+	formattedResponseStr := string(formattedResponse)
+	// debug log -- to be removed at a later stage
+	log.Printf("The formatted response: %s", formattedResponseStr)
```
Liam-Doodson: Should we remove this now. Better to be safe on the possibility of logging sensitive info!

```suggestion
```

.gitignore
keremgocen: I think it makes sense to keep them and not ignore, in the previous projects when I used mockgen that was the case as well. For some reason this got ignored along with the binary earlier, I'll fix it.

README.md
```diff
@@ -49,7 +49,7 @@ Create or update your VSCode MCP configuration file (`mcp.json`), as document he
   "servers": {
     "neo4j": {
       "type": "stdio",
-      "command": "neo4j-mcp",
+      "command": "./bin/neo4j-mcp", // Use full path to binary or ensure neo4j-mcp is in PATH
```
MacondoExpress: As we don't have an user facing documentation, I would do the other way around similar as it is done in aura-cli https://neo4j.com/docs/aura/aura-cli/installation/.
And then move the "build from source" at the end of the README or to a DEVELOPMENT file.

README.md
```diff
@@ -65,6 +65,19 @@ Adjust the environment variables according to your Neo4j instance configuration.
 
 Open the VSCode chat in agentic mode and ask about your configured Neo4j database.
 
+## Documentation
+
+📚 **[Development Guide](docs/README.md)** - Testing, mockgen setup, and development workflows
+
+## Logging
+
+This project follows the [MCP specification](https://spec.modelcontextprotocol.io/specification/basic/transports/#stdio) recommendation that all log output should be written to **stderr** to keep **stdout** clean for protocol communication.
```
MacondoExpress: The link seems to not redirect where the recommendation is present. I found the recommendation here: https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#stdio
I think we should also specify that the recommendation is mainly aimed as the STDIO version

README.md
```diff
@@ -65,6 +65,19 @@ Adjust the environment variables according to your Neo4j instance configuration.
...
+
+We achieve this by using Go's standard `log` package, which writes to stderr by default. This ensures:
```
MacondoExpress, line 76: Very-nitpick: I think we could be more specific here, for instance: we're using the `default` or `standard` logger from the `log` package, which writes to stderr by default.
From the log package documentation:
_It also has a predefined 'standard' Logger accessible through helper functions Print[f|ln], Fatal[f|ln], and Panic[f|ln], which are easier to use than creating a Logger manually. That logger writes to standard error_

docs/README.md
```diff
@@ -0,0 +1,56 @@
...
+        ExecuteReadQuery(gomock.Any(), "MATCH (n) RETURN n", gomock.Nil(), "neo4j").
+        Return([]*neo4j.Record{}, nil)
+
+    // Use mockDB in your test...
+}
+```
+
+See `internal/tools/get_schema_handler_gomock_test.go` for complete examples.
+
+## Project Structure
```
MacondoExpress, line 48: I would avoid documenting the current Project structure as it is still to be properly defined

internal/tools/get_schema_handler.go
```diff
@@ -2,43 +2,57 @@ package tools
...
-		return handleGetSchema(ctx, deps.Driver, deps.Config)
+		return handleGetSchema(ctx, deps.DBService, deps.Config)
 	}
 }
 
-func handleGetSchema(ctx context.Context, driver *neo4j.DriverWithContext, config *config.Config) (*mcp.CallToolResult, error) {
-	fmt.Fprintf(os.Stderr, "Retrieving Neo4j schema information using APOC\n")
+// handleGetSchema retrieves Neo4j schema information using APOC
+func handleGetSchema(ctx context.Context, dbService database.DatabaseService, config *config.Config) (*mcp.CallToolResult, error) {
+	if dbService == nil {
```
MacondoExpress, line 32: I think these errors will be better raised outside the specific tool invocation.
I think that it will be better raise these in `RegisterAllTools` outside the `ToolResultError` wrapper

docs/README.md
```diff
@@ -0,0 +1,56 @@
+# Neo4j MCP Development Guide
```
MacondoExpress, line 1: Fan of this doc! I think is worth changing the name to better reflect the intention of the file and why it's different from the root README file.

internal/server/server.go
MacondoExpress: Nice!

internal/database/service.go
```diff
@@ -0,0 +1,80 @@
...
+
+// NewNeo4jService creates a new Neo4jService instance
+func NewNeo4jService(driver *neo4j.DriverWithContext) DatabaseService {
+	return &Neo4jService{
+		driver: driver,
+	}
+}
+
+// ExecuteReadQuery executes a read-only Cypher query and returns raw records
+func (s *Neo4jService) ExecuteReadQuery(ctx context.Context, cypher string, params map[string]any, database string) ([]*neo4j.Record, error) {
```
MacondoExpress, line 25: I think we can remove the execute_query file and neo4j_records_to_json

docs/README.md
keremgocen: we will address the doc comments in a separate PR

PR #6, opened 2025-09-11
Add test coverage for server, config and tooling

keremgocen:
[trello](https://trello.com/c/YdzVLzYI/2313-add-simple-tests-to-mcp)

- [x] adds test coverage for server, RunCypherHandler
- [x] config tests and validation
- [x] simplifies GetSchemaHandler tests (it was a bit too complex for what we have now)
~- [ ] add coverage for database service (maybe make neo4j a dependency injection instead of a connection string)~

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -26,7 +26,6 @@ func (c *Config) Validate() error {
 		{c.URI, "Neo4j URI"},
 		{c.Username, "Neo4j username"},
 		{c.Password, "Neo4j password"},
-		{c.Database, "Neo4j database name"},
 	}
 
 	for _, v := range validations {
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -62,15 +62,15 @@ func TestConfig_Validate(t *testing.T) {
 			errMsg:  "Neo4j password is required but was empty",
 		},
 		{
-			name: "empty database",
+			name: "empty database should not raise error",
 			cfg: &Config{
 				URI:      "bolt://localhost:7687",
 				Username: "neo4j",
 				Password: "password",
 				Database: "",
 			},
-			wantErr: true,
-			errMsg:  "Neo4j database name is required but was empty",
+			wantErr: false,
+			errMsg:  "",
 		},
 	}
 


internal/server/server.go
```diff
@@ -20,9 +20,10 @@ type Neo4jMCPServer struct {
 }
 
 // NewNeo4jMCPServer creates a new MCP server instance
+// The config parameter is expected to be already validated
 func NewNeo4jMCPServer(cfg *config.Config) (*Neo4jMCPServer, error) {
 	mcpServer := server.NewMCPServer(
-		"neo4-mcp",
+		"neo4j-mcp",
```
MacondoExpress, line 26: ❤️

internal/tools/get_schema_handler_test.go
```diff
@@ -14,160 +13,43 @@ import (
...
-		})
-	}
-}
-
-func TestHandleGetSchema(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 
-	t.Run("successful execution with specific assertions", func(t *testing.T) {
+	t.Run("successful schema retrieval", func(t *testing.T) {
```
MacondoExpress, line 19: 🚀  really like this approach compared with the initial one!

internal/tools/get_schema_handler_test.go
```diff
@@ -197,9 +83,37 @@ func TestGetSchemaHandler_InputValidation(t *testing.T) {
...
+			t.Errorf("Expected no error from handler, got: %v", err)
+		}
 		if result == nil || !result.IsError {
-			t.Error("Expected error result for nil config")
+			t.Error("Expected error result for nil database service")
 		}
 	})
 }
+
+func TestGetSchemaSpec(t *testing.T) {
```
MacondoExpress: I would keep it in a separate file this, but I think we can test the spec directly from the `RegisterAllTools` rather than having a test file for spec

internal/tools/run_cypher_handler_test.go
```diff
@@ -0,0 +1,281 @@
...
+		deps := &ToolDependencies{
+			Config:    &config.Config{Database: "testdb"},
+			DBService: mockDB,
+		}
+
+		handler := RunCypherHandler(deps)
+		request := mcp.CallToolRequest{
+			Params: mcp.CallToolParams{
+				Arguments: map[string]any{
+					"query": "MATCH (n) RETURN count(n)",
```
MacondoExpress, line 71: You can separate this query in a variable and reuse it in the mock

internal/tools/get_schema_handler_test.go
keremgocen: removed

PR #7, opened 2025-09-12
Add a simple CI configuration

keremgocen:
- introduce a basic continuous integration setup for building and testing the project using GitHub Actions.
- update Go version and dependencies to their latest versions.

(maybe add a basic .golangci-lint config as well)

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -4,8 +4,7 @@ on:
   push:
     branches: [main]
   pull_request:
-    # Uncomment this line to run CI only against main branch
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,45 @@
+name: Build and Test
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    # Uncomment this line to run CI only against main branch
+    # branches: [main]
```
keremgocen: uncomment this before merging so we only run it for PRs against the main branch

PR #32, opened 2025-09-18
Update documentation and contributing guidelines

keremgocen:
Enhance error handling documentation, provide clearer usage examples, and streamline the README for better clarity. Update contributing guidelines to improve the onboarding process for new contributors.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,7 +88,7 @@ func TestMyFunction(t *testing.T) {
 }
 ```
 
-See `internal/tools/get_schema_handler_gomock_test.go` for a fuller pattern.
+See `internal/tools/get_schema_handler_test.go` for a fuller pattern.
 
 ## Testing using the @modelcontextprotocol/inspector:
 


CONTRIBUTING.md
```diff
@@ -0,0 +1,158 @@
+# Contributing to Neo4j MCP
+
+Thank you for your interest in contributing to the Neo4j MCP server! This document provides guidelines and information for contributors.
+
+## Code of Conduct
+
+This project follows the [Neo4j Community Guidelines](todo: missing link here). Please read and follow these guidelines to ensure a welcoming environment for everyone.
```
keremgocen: we have a bunch of todos throughout, which I intend to remove.

do we have a Neo4j code of conduct link we can use?

README.md
```diff
@@ -61,51 +58,47 @@ Create or update your VSCode MCP configuration file (`mcp.json`), as document he
 }
 ```
 
-Adjust the environment variables according to your Neo4j instance configuration.
+<!-- TODO: add claude desktop MCP installation instructions -->
```
keremgocen: github-mcp-server does it nicely here, should we add similar links? https://github.com/github/github-mcp-server?tab=readme-ov-file#install-in-other-mcp-hosts

CONTRIBUTING.md
LackOfMorals: Any who wants to make contribution needs to sign up here:- https://neo4j.com/developer/contributing-code/#sign-cla  before they can do so.  We need to add that in.

CONTRIBUTING.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Project Status
+
+Active development; not yet production‑hardened.
```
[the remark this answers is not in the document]
LackOfMorals: Suggest we phrase as

Active development; not yet suitable for production

README.md
LackOfMorals, line 1: With the Aura CLI, the README.md  describes what the Aura CLI does, how to use it and then directs anyone who wishes to contribute to read the CONTRIBUTING. md doc. 

I would be tempted to follow that format and keep the README.md primarily on the topics of what this project is and how to get started using it.

keremgocen: Currently, the README.md is targeted for users while CONTRIBUTING.md is more for the developers. Is there anything specific we need to address? Otherwise I think this is already what we have.

README.md
```diff
@@ -1,55 +1,55 @@
...
-This project is currently under active development and is not ready for production use.
+Active development; not yet suitable for production
 
-## Setup
+## Prerequisites
 
-### 1. Clone the Repository
+- A running Neo4j database instance; either local [neo4j–desktop](https://neo4j.com/download/) or [Aura](https://neo4j.com/product/auradb/).
+- APOC plugin installed in the Neo4j instance.
+- Any MCP-compatible client (ie. [VSCode](https://code.visualstudio.com/) with [MCP support](https://code.visualstudio.com/docs/copilot/customization/mcp-servers))
```
Liam-Doodson: I think this should be e.g. to make clear it's just one example. It might also be good to suggest alternatives like claude

```suggestion
- Any MCP-compatible client (e.g. [VSCode](https://code.visualstudio.com/) with [MCP support](https://code.visualstudio.com/docs/copilot/customization/mcp-servers))
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,154 @@
...
+git clone git@github.com:neo4j/mcp.git && cd mcp
+```
+
+## Install Dependencies
+
+```bash
+# Install Go dependencies
+go mod download
+
+# Install mock generator (only if you will change interfaces)
```
mjfwebb: What does change interfaces mean?

I think it would be beneficial to expand on this a little bit.

CONTRIBUTING.md
```diff
@@ -0,0 +1,154 @@
...
+go test ./... -cover
+
+# Verbose / single package
+go test ./internal/tools -v
+
+# Build binary
+go build -C cmd/neo4j-mcp -o ../../bin/
+
+# Run from source
+go run ./cmd/neo4j-mcp
```
mjfwebb, line 59: I'm being intentionally dumb, and just following the commands.

Apparently I needed to have the database running at this point:

➜  mcp git:(main) go run ./cmd/neo4j-mcp
2025/09/23 10:37:06 Starting Neo4j MCP Server...
2025/09/23 10:37:06 Server error: failed to verify database connectivity: ConnectivityError: dial tcp 127.0.0.1:7687: connect: connection refused
exit status 1


Also, is this the point in which I need env? Maybe worth pointing it out

CONTRIBUTING.md
```diff
@@ -0,0 +1,154 @@
...
+
+# Optional: install
+go install -C cmd/neo4j-mcp
```
mjfwebb: What does this do?

CONTRIBUTING.md
```diff
@@ -0,0 +1,154 @@
...
+    mockDB := mocks.NewMockDatabaseService(ctrl)
+    mockDB.EXPECT().
+        ExecuteReadQuery(gomock.Any(), "MATCH (n) RETURN n", gomock.Nil(), "neo4j").
+        Return([]*neo4j.Record{}, nil)
+
+    // Use mockDB in your test ...
+}
+```
+
+See `internal/tools/get_schema_handler_gomock_test.go` for a fuller pattern.
```
mjfwebb: This file doesn't exist

CONTRIBUTING.md
```diff
@@ -0,0 +1,154 @@
...
+
+The Neo4j MCP capabilities can be tested using the `@modelcontextprotocol/inspector`:
+
+```bash
+npx @modelcontextprotocol/inspector go run ./cmd/neo4j-mcp
+```
+
+## Adding New MCP Tools
+
+1. **Define tool specification** in `internal/tools/`:
```
mjfwebb:
```suggestion
1. **Define tool specifications** in `internal/tools/`:
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,154 @@
...
+## Environment Variables
+
+For local testing, set these environment variables:
+
+```bash
+export NEO4J_URI="bolt://localhost:7687"
+export NEO4J_USERNAME="neo4j"
+export NEO4J_PASSWORD="password"
+export NEO4J_DATABASE="neo4j"
+```
```
mjfwebb: Could be good to mention the default values are as written, so setting environment variables are only required if you want to change them.

CONTRIBUTING.md
keremgocen: we have some interface defined here https://github.com/neo4j/mcp/blob/main/internal/database/interfaces.go, that's it for now but we might add more in the future, they describe the behaviour of the database service in this case.

They are also used to generate mocks so we can use them in tests, without needing a database instance running. So when they are modified, mocks need to be generated again. I'll try to make dependency that more clear.

CONTRIBUTING.md
keremgocen: yes, before we run the binary we need the env variable setup. And the database dependency should be already mentioned in the README as a prerequisite.

CONTRIBUTING.md
keremgocen: this is an install command for the binary in  `cmd/neo4j-mcp` folder,`-C` changes the directory to that folder. SO this is a bit brittle that if you're not in the `mcp` root directory the command will fail. I think we should mention that as well.

CONTRIBUTING.md
keremgocen: good catch, it was renamed to `internal/tools/get_schema_handler_test.go`

CONTRIBUTING.md
keremgocen: normally I would not even set the default values at all, but this looks like a pretty standard way of using the Neo4j Desktop client so maybe that's ok? I'm voting for having a .env-template file and putting the default values in there. What do you think? (cc: @MacondoExpress)

CONTRIBUTING.md
keremgocen: updated the wording to make the dependency more obvious

CONTRIBUTING.md
keremgocen: added run from root

CONTRIBUTING.md
keremgocen: updated file name

CONTRIBUTING.md
keremgocen: moved env variables up before this section

PR #33, opened 2025-09-18
Added `-v` flag to check current version in use

MacondoExpress:
Running `neo4j-mcp -v` now returns `neo4j-mcp version: {{current version}}`

### DIFF
--- a/cmd/neo4j-mcp/main.go
+++ b/cmd/neo4j-mcp/main.go
@@ -2,13 +2,21 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"log"
+	"os"
 
 	"github.com/neo4j/mcp/internal/config"
 	"github.com/neo4j/mcp/internal/server"
 )
 
 func main() {
+	// Handle version flag
+	if len(os.Args) > 1 && os.Args[1] == "-v" {
+		// NOTE: "standard" log package logger write on on STDERR, in this case we want explicitly to write to STDOUT
+		fmt.Printf("neo4j-mcp version: %s\n", server.Version)
+		return
+	}
 	// get config from environment variables
 	cfg, err := config.LoadConfig()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/server/server.go
```diff
@@ -12,19 +12,23 @@ import (
 	"github.com/neo4j/neo4j-go-driver/v5/neo4j"
 )
 
+// Version represents the current version of the neo4j-mcp server
+const Version = "0.1.1"
```
keremgocen, line 16: Nit: I’m thinking if this should come from the config instead of a constant defined server.go? Not sure how it is done in aura-cli or in graphql library.

cmd/neo4j-mcp/main.go
```diff
@@ -2,13 +2,21 @@ package main
...
 
 	"github.com/neo4j/mcp/internal/config"
 	"github.com/neo4j/mcp/internal/server"
 )
 
 func main() {
+	// Handle version flag
+	if len(os.Args) > 1 && os.Args[1] == "-v" {
+		// NOTE: "standard" log package logger write on on STDERR, in this case we want explicitly to write to STDOUT
+		fmt.Printf("neo4j-mcp version: %s\n", server.Version)
```
keremgocen, line 17: I’m not against using fmt (especially in main) but just so we are aware the logger output can be specified in the log package as well. 
Not a blocker since we are already considering to add a structured logger in the feature. Even then we might want to keep the “fmt” stuff in main.

MacondoExpress: I thought it was more verbose to change the standard logger then print, but happy to change it, but as you said, we will revisit the whole logging mechanisms soon!

internal/server/server.go
MacondoExpress: Not 100% sure if I'm happy as well with the server constant, but I would not put this in the Config as is not a configurable item IMO!

PR #52, opened 2025-09-30
Enhance Neo4j service with error handling and testing improvements

keremgocen:
Introduce error handling in the Neo4j service, implement a session interface for better testability, and add realistic test scenarios. Refactor code to improve clarity and maintainability.

- expands the interfaces we are using from Neo4j driver for testability, they are now moved into adapters.go to keep it tidy
- more unit test coverage in service.go now but ideally we'd like to introduce integration tests since not everything in the driver is exported so it's hard to test in unit tests (you need to do a lot of mocking and wrapping)

[trello](https://trello.com/c/p4bxHuIb/2348-mcp-database-service-test-coverage-driver-mocks)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -100,34 +100,53 @@ npx @modelcontextprotocol/inspector go run ./cmd/neo4j-mcp
 
 ## MCP Error Handling
 
-**Note:** MCP error handling differs from standard Go patterns. Instead of returning Go errors directly, we wrap error information inside a `CallToolResult` with the `IsError` flag set to true to properly communicate MCP-related errors to clients.
+MCP error handling follows a specific pattern that differs from standard Go error handling. According to the [MCP specification](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling), tool handlers should communicate errors through the tool result structure rather than returning Go errors directly.
 
-**MCP Tool Handler error handling pattern:**
-When implementing MCP tool handlers, return errors using the MCP tool result structure:
+### When to use MCP tool result errors vs direct Go errors:
+
+- **Use MCP tool result errors** (`NewToolResultError`) for:
+
+  - Business logic errors (invalid input, database constraints, etc.)
+  - Operational errors that the client should handle gracefully
+  - Any error that represents a meaningful response to the client
+
+- **Return Go errors directly** for:
+  - System-level failures (out of memory, network failures)
+  - Programming errors that indicate bugs in the server implementation
+  - Cases where the server cannot continue processing
+
+### Recommended MCP Tool Handler error handling pattern:
+
+When implementing MCP tool handlers, use the `mcp.NewToolResultError` helper function for cleaner error handling:
 
 ```go
 func MyToolHandler(deps *ToolDependencies) mcp.ToolHandler {
     return func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-        // ... implementation ...
+        // Bind and validate arguments
+        var args MyToolInput
+        if err := request.BindArguments(&args); err != nil {
+            return mcp.NewToolResultError("Invalid arguments: " + err.Error()), nil
+        }
 
+        // Business logic validation
+        if args.SomeField == "" {
+            return mcp.NewToolResultError("SomeField is required"), nil
+        }
+
+        // Execute operation
+        result, err := someOperation(ctx, args)
         if err != nil {
-            return &mcp.CallToolResult{
-                Content: []mcp.Content{
-                    mcp.NewTextContent(fmt.Sprintf("Error: %v", err)),
-                },
-                IsError: &[]bool{true}[0],
-            }, nil // Note: return nil as the second parameter, error info is in the result
+            // Use MCP error for business/operational errors
+            return mcp.NewToolResultError("Operation failed: " + err.Error()), nil
         }
 
         // Success case
-        return &mcp.CallToolResult{
-            Content: []mcp.Content{
-                mcp.NewTextContent("Success result"),
-            },
-        }, nil
+        return mcp.NewToolResultText(result), nil
     }
 }
 ```
+
+**Note:** Always return `nil` as the second parameter when using `NewToolResultError`, as the error information is embedded within the `CallToolResult` structure.
 
 ## Adding New MCP Tools
 


CONTRIBUTING.md
```diff
@@ -98,6 +98,37 @@ The Neo4j MCP capabilities can be tested using the `@modelcontextprotocol/inspec
 npx @modelcontextprotocol/inspector go run ./cmd/neo4j-mcp
 ```
 
+## MCP Error Handling
```
keremgocen, line 101: sneaked in the error handling approach for MCP, happy to move this to a separate PR

MacondoExpress: Nice!

CONTRIBUTING.md
```diff
@@ -98,6 +98,37 @@ The Neo4j MCP capabilities can be tested using the `@modelcontextprotocol/inspec
...
+
+**Note:** MCP error handling differs from standard Go patterns. Instead of returning Go errors directly, we wrap error information inside a `CallToolResult` with the `IsError` flag set to true to properly communicate MCP-related errors to clients.
```
MacondoExpress: I think we should document the reason why this is required, and when we should return the error directly instead.
A link to the tool error handling could be beneficial:
[https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling)

We can have a more straightforward example using: `NewToolResultError` from `mcp-go` instead of manually wrap the error ourself in the CallToolResult

internal/database/adapters.go
```diff
@@ -0,0 +1,44 @@
...
+
+// ExecuteWrite executes a write transaction
+func (a *neo4jSessionAdapter) ExecuteWrite(ctx context.Context, work func(tx neo4j.ManagedTransaction) (any, error)) (any, error) {
+	return a.session.ExecuteWrite(ctx, work)
+}
+
+// Close closes the session
+func (a *neo4jSessionAdapter) Close(ctx context.Context) error {
+	return a.session.Close(ctx)
+}
```
MacondoExpress: no new line

cmd/neo4j-mcp/main.go
```diff
@@ -25,14 +27,35 @@ func main() {
...
 	// Gracefully handle shutdown
 	ctx := context.Background()
+	defer func() {
+		if err := driver.Close(ctx); err != nil {
+			log.Fatalf("Error closing driver: %v", err)
+		}
+	}()
+
+	// Verify database connectivity
+	if err := driver.VerifyConnectivity(ctx); err != nil {
```
keremgocen, line 45: We move the driver to the main, which helps create a cleaner interface on the internals, and we can still drop in a mock driver for unit tests.

internal/database/service.go
```diff
@@ -11,26 +11,24 @@ import (
 
 // Neo4jService is the concrete implementation of DatabaseService
 type Neo4jService struct {
-	driver *neo4j.DriverWithContext
+	driver neo4j.DriverWithContext
```
keremgocen, line 15: This was just a pointer to the interface, so I removed the pointer notation, which is not needed

internal/database/service.go
```diff
@@ -11,26 +11,24 @@ import (
...
 }
 
 // NewNeo4jService creates a new Neo4jService instance
-func NewNeo4jService(driver *neo4j.DriverWithContext) DatabaseService {
+func NewNeo4jService(driver neo4j.DriverWithContext) (DatabaseService, error) {
+	if driver == nil {
```
keremgocen, line 20: The service is not responsible anymore for creating the driver now, but it checks if it's missing or not.

internal/database/service.go
```diff
@@ -11,26 +11,24 @@ import (
...
 func (s *Neo4jService) ExecuteReadQuery(ctx context.Context, cypher string, params map[string]any, database string) ([]*neo4j.Record, error) {
-	if s.driver == nil {
-		err := fmt.Errorf("Neo4j driver is not initialized")
-		log.Printf("Error in ExecuteReadQuery: %v", err)
-		return nil, err
-	}
-
-	res, err := neo4j.ExecuteQuery(ctx, *s.driver, cypher, params, neo4j.EagerResultTransformer, neo4j.ExecuteQueryWithDatabase(database), neo4j.ExecuteQueryWithReadersRouting())
 
+	res, err := neo4j.ExecuteQuery(ctx, s.driver, cypher, params, neo4j.EagerResultTransformer, neo4j.ExecuteQueryWithDatabase(database), neo4j.ExecuteQueryWithReadersRouting())
```
keremgocen, line 32: This is the bit we can't test in our unit tests, because we are calling the `neo4j.ExecuteQuery` directly, and I could not find a good way to mock it or convert to a dependency unless we use the exported https://github.com/neo4j/neo4j-go-driver/blob/6.x/neo4j/session.go#L57 functions, which requires us to manage our own sessions. Maybe in the future, if we want to test anything else inside these execution functions, we can go that way.

internal/server/server.go
```diff
@@ -31,53 +29,41 @@ func NewNeo4jMCPServer(version string, cfg *config.Config) (*Neo4jMCPServer, err
...
 	log.Println("Started Neo4j MCP Server. Now listening for input...")
+	// Note: ServeStdio handles its own signal management for graceful shutdown
 	return server.ServeStdio(s.mcpServer)
 }
 
-// Stop gracefully stops the server and closes the driver
-func (s *Neo4jMCPServer) Stop(ctx context.Context) error {
-	return (*s.driver).Close(ctx)
+// Stop gracefully stops the server
+func (s *Neo4jMCPServer) Stop() error {
```
keremgocen, line 64: the context in Start()/Stop() and handler contexts are different:

Start context: Created in main.go, it was used only for startup connectivity verification
Handler contexts: Created by MCP library per-request, used for actual database operations

since Start() and Stop() functions of the server only use context for one database connectivity check (which we already do in main.go), I just removed it.

go.mod
```diff
@@ -13,8 +13,8 @@ require (
 	github.com/buger/jsonparser v1.1.1 // indirect
 	github.com/google/uuid v1.6.0 // indirect
 	github.com/invopop/jsonschema v0.13.0 // indirect
-	github.com/mailru/easyjson v0.9.0 // indirect
-	github.com/spf13/cast v1.10.0 // indirect
```
mjfwebb, line 17: Is there a particular reason we're downgrading these?

keremgocen: I was not aware, I’ll do another go mod tidy

CONTRIBUTING.md
keremgocen: Good call, I’ll update the docs with that

go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/buger/jsonparser v1.1.1 // indirect
 	github.com/google/uuid v1.6.0 // indirect
 	github.com/invopop/jsonschema v0.13.0 // indirect
-	github.com/mailru/easyjson v0.9.0 // indirect
+	github.com/mailru/easyjson v0.9.1 // indirect
```
keremgocen, line 16: looks like I had to do `go get -u all` do upgrade these packages from what they were and now it's even a bit ahead:
```
go: downloading github.com/mailru/easyjson v0.9.1
go: upgraded github.com/mailru/easyjson v0.7.7 => v0.9.1
go: upgraded github.com/spf13/cast v1.7.1 => v1.10.0
```

CONTRIBUTING.md
keremgocen: updated the docs as well

PR #53, opened 2025-09-30
Introduce read-cypher (read-only) tool and rename run-cypher to write-cypher

MacondoExpress:
Introduce read-cypher and enforce read-only execution by classifying queries with GetQueryType (via EXPLAIN) before running them.

Implementation logic:
- Calls GetQueryType; if it errors, returns a ToolResultError and does not execute the query.
- Rejects any query not classified as "r" and directs the user to use write-cypher for write operations, schema or admin commands, or PROFILE queries.

Some of the above behaviors could cause confusion. For example:
- Admin queries that do not mutate the database/DBMS (e.g., SHOW USERS) are treated as non-read queries.
- EXPLAIN PROFILE queries are treated as non-read queries, even if the statements following "EXPLAIN PROFILE" do not mutate the database.

### DIFF
--- a/internal/database/service.go
+++ b/internal/database/service.go
@@ -59,6 +59,48 @@ func (s *Neo4jService) ExecuteWriteQuery(ctx context.Context, cypher string, par
 	return res.Records, nil
 }
 
+
+// GetQueryType prefixes the provided query with EXPLAIN and returns the query type (e.g. 'r' for read, 'w' for write, 'rw' etc.)
+// This allows read-only tools to determine if a query is safe to run in read-only context.
+func (s *Neo4jService) GetQueryType(ctx context.Context, cypher string, params map[string]any, database string) (string, error) {
+	if s.driver == nil {
+		err := fmt.Errorf("Neo4j driver is not initialized")
+		log.Printf("Error in GetQueryType: %v", err)
+		return "", err
+	}
+
+	explain := "EXPLAIN " + cypher
+	res, err := neo4j.ExecuteQuery(ctx, *s.driver, explain, params, neo4j.EagerResultTransformer, neo4j.ExecuteQueryWithDatabase(database))
+	if err != nil {
+		wrappedErr := fmt.Errorf("Error during GetQueryType: %w", err)
+		log.Printf("Error during GetQueryType:: %v", wrappedErr)
+		return "", wrappedErr
+	}
+
+	if res.Summary == nil {
+		err := fmt.Errorf("Error during GetQueryType: no summary returned for explained query")
+		log.Printf("Error during GetQueryType:: %v", err)
+		return "", err
+	}
+
+	st := res.Summary.StatementType()
+	var qType string
+	switch st {
+	case neo4j.StatementTypeReadOnly:
+		qType = "r"
+	case neo4j.StatementTypeReadWrite:
+		qType = "rw"
+	case neo4j.StatementTypeWriteOnly:
+		qType = "w"
+	case neo4j.StatementTypeSchemaWrite:
+		qType = "s"
+	default:
+		qType = "unknown"
+	}
+
+	return qType, nil
+}
+
 // Neo4jRecordsToJSON converts Neo4j records to JSON string
 func (s *Neo4jService) Neo4jRecordsToJSON(records []*neo4j.Record) (string, error) {
 	results := make([]map[string]any, 0)
--- a/internal/database/service.go
+++ b/internal/database/service.go
@@ -76,6 +76,5 @@ func (s *Neo4jService) Neo4jRecordsToJSON(records []*neo4j.Record) (string, erro
 
 	formattedResponseStr := string(formattedResponse)
 
-
 	return formattedResponseStr, nil
 }
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -31,7 +31,7 @@ func NewNeo4jMCPServer(cfg *config.Config) (*Neo4jMCPServer, error) {
 		Version,
 		server.WithToolCapabilities(true),
 		server.WithInstructions("This is the Neo4j official MCP server and can provide tool calling to interact with your Neo4j database,"+
-			"by inferring the schema with tools like get-schema and executing arbitrary Cypher queries with write-cypher."),
+			"by inferring the schema with tools like get-schema and executing arbitrary Cypher queries with read-cypher."),
 	)
 
 	// Initialize Neo4j driver once
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -31,7 +31,7 @@ func NewNeo4jMCPServer(cfg *config.Config) (*Neo4jMCPServer, error) {
 		Version,
 		server.WithToolCapabilities(true),
 		server.WithInstructions("This is the Neo4j official MCP server and can provide tool calling to interact with your Neo4j database,"+
-			"by inferring the schema with tools like get-schema and executing arbitrary Cypher queries with run-cypher."),
+			"by inferring the schema with tools like get-schema and executing arbitrary Cypher queries with write-cypher."),
 	)
 
 	// Initialize Neo4j driver once
--- a/internal/tools/read_cypher_handler.go
+++ b/internal/tools/read_cypher_handler.go
@@ -39,12 +39,21 @@ func handleReadCypher(ctx context.Context, request mcp.CallToolRequest, dbServic
 		log.Printf("%s", errMessage)
 		return mcp.NewToolResultError(errMessage), nil
 	}
-	// if strings.ContainsAny(strings.ToLower(Query), "explain") {
-	// 	// records, err := dbService.ExecuteReadQuery(ctx, Query, Params, config.Database)
-	// 	// log.Print(records)
-	// 	// log.Print(err)
-	// }
-	// Execute the Cypher query using the database service
+
+	// Get queryType by pre-appending "EXPLAIN" to identify if the query is of type "r", if not raise a ToolResultError
+	queryType, err := dbService.GetQueryType(ctx, Query, Params, config.Database)
+	if err != nil {
+		log.Printf("Error while classifying Cypher query: %v", err)
+		return mcp.NewToolResultError(err.Error()), nil
+	}
+
+	if queryType != "r" { // only queryType == "r" are allowed in read-cypher
+		errMessage := "read-cypher can run only read-only Cypher statements. For write operations (CREATE, MERGE, DELETE, SET, etc...), schema/admin commands, or PROFILE queries, use write-cypher instead."
+		log.Printf("Rejected non-read query (type=%s): %s", queryType, Query)
+		return mcp.NewToolResultError(errMessage), nil
+	}
+
+	// Execute the Cypher query using the database service (now confirmed read-only)
 	records, err := dbService.ExecuteReadQuery(ctx, Query, Params, config.Database)
 	if err != nil {
 		log.Printf("Error executing Cypher query: %v", err)
--- a/internal/tools/read_cypher_handler.go
+++ b/internal/tools/read_cypher_handler.go
@@ -1,0 +1,61 @@
+package tools
+
+import (
+	"context"
+	"log"
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/neo4j/mcp/internal/config"
+	"github.com/neo4j/mcp/internal/database"
+)
+
+func ReadCypherHandler(deps *ToolDependencies) func(context.Context, mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+	return func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+		return handleReadCypher(ctx, request, deps.DBService, deps.Config)
+	}
+}
+
+func handleReadCypher(ctx context.Context, request mcp.CallToolRequest, dbService database.DatabaseService, config *config.Config) (*mcp.CallToolResult, error) {
+	var args ReadCypherInput
+	// Bind arguments to the struct
+	if err := request.BindArguments(&args); err != nil {
+		log.Printf("Error binding arguments: %v", err)
+		return mcp.NewToolResultError(err.Error()), nil
+	}
+	Query := args.Query
+	Params := args.Params
+
+	log.Printf("cypher-query: %s", Query)
+
+	// Validate that query is not empty
+	if Query == "" {
+		errMessage := "Query parameter is required and cannot be empty"
+		log.Printf("%s", errMessage)
+		return mcp.NewToolResultError(errMessage), nil
+	}
+
+	if dbService == nil {
+		errMessage := "Database service is not initialized"
+		log.Printf("%s", errMessage)
+		return mcp.NewToolResultError(errMessage), nil
+	}
+	// if strings.ContainsAny(strings.ToLower(Query), "explain") {
+	// 	// records, err := dbService.ExecuteReadQuery(ctx, Query, Params, config.Database)
+	// 	// log.Print(records)
+	// 	// log.Print(err)
+	// }
+	// Execute the Cypher query using the database service
+	records, err := dbService.ExecuteReadQuery(ctx, Query, Params, config.Database)
+	if err != nil {
+		log.Printf("Error executing Cypher query: %v", err)
+		return mcp.NewToolResultError(err.Error()), nil
+	}
+
+	response, err := dbService.Neo4jRecordsToJSON(records)
+	if err != nil {
+		log.Printf("Error formatting query results: %v", err)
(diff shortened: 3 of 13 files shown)


internal/tools/write_cypher_spec.go
```diff
@@ -4,16 +4,16 @@ import (
...
 }
 
-func RunCypherSpec() mcp.Tool {
-	return mcp.NewTool("run-cypher",
-		mcp.WithDescription("run-cypher executes any arbitrary Cypher query against the user-configured Neo4j database."),
-		mcp.WithInputSchema[RunCypherInput](),
-		mcp.WithTitleAnnotation("Run Cypher"),
+func WriteCypherSpec() mcp.Tool {
+	return mcp.NewTool("write-cypher",
+		mcp.WithDescription("write-cypher executes any arbitrary Cypher query, with write access, against the user-configured Neo4j database."),
```
Liam-Doodson, line 14: Not sure if this is the place but is it worth telling the LLM to prefer read-cypher if it's available? We could add something to the description like "If the query does not contain any write operations, schema/admin commands, or PROFILE queries, use read-cypher instead."

MacondoExpress: Not sure, I feel that for the `read-cypher` it makes sense as it helps understand when to use the tool and a potential error returned, for the `write-cypher` I'm more unsure about if we deliver some value or if degrade the experience by redirecting the LLM to `read-cypher`. Mainly, because I feel that use case could be solved by a `readonly` mode. But, I'm honestly unsure. Maybe slightly leaning on keep it the description simple.

Liam-Doodson: Yeah you definitely make a good point!

My thinking is that we don't want to overload the user with approvals on the write-cypher tool as it makes it much more likely they click the "Always Allow" option in their agent and then let something slip through they didn't want to

I don't feel that strongly though and, as you say, maybe users that are concerned about what "write-cypher" can do should just be using `readonly` anyway (once we add that option)

MacondoExpress: I'm open to change it if you feel otherwise, even in any other direction: grammatically, phrasing etc ...

internal/tools/read_cypher_handler.go
```diff
@@ -0,0 +1,70 @@
...
+	var args ReadCypherInput
+	// Bind arguments to the struct
+	if err := request.BindArguments(&args); err != nil {
+		log.Printf("Error binding arguments: %v", err)
+		return mcp.NewToolResultError(err.Error()), nil
+	}
+	Query := args.Query
+	Params := args.Params
+
+	log.Printf("cypher-query: %s", Query)
```
keremgocen, line 29: was this left for debugging purposes?

internal/tools/write_cypher_spec.go
keremgocen: I'm also curious what would happen if we add a prompt to tell the LLM to use write-cypher for PROFILE and SHOW USERS, would that be consistent enough to rely on?
We mentioned in the description:
>Admin queries that do not mutate the database/DBMS (e.g., SHOW USERS) are treated as non-read queries.
EXPLAIN PROFILE queries are treated as non-read queries, even if the statements following "EXPLAIN PROFILE" do not mutate the database.

But I don't feel like we are making it clear in the README, or it could help for users to mention it even in the code as a comment where the validation happens, or in the spec if that works.

internal/server/server.go
```diff
@@ -28,7 +28,7 @@ func NewNeo4jMCPServer(version string, cfg *config.Config) (*Neo4jMCPServer, err
 		version,
 		server.WithToolCapabilities(true),
 		server.WithInstructions("This is the Neo4j official MCP server and can provide tool calling to interact with your Neo4j database,"+
-			"by inferring the schema with tools like get-schema and executing arbitrary Cypher queries with run-cypher."),
+			"by inferring the schema with tools like get-schema and executing arbitrary Cypher queries with read-cypher."),
```
keremgocen, line 31:
```suggestion
			"by inferring the schema with tools like get-schema and executing arbitrary Cypher queries with read-cypher or write-cypher."),
```
server description only mentions read-cypher, we should add both tools, I think

internal/tools/run_cypher_handler_test.go
```diff
@@ -258,24 +258,3 @@ func TestRunCypherHandler(t *testing.T) {
 	})
 }
 
-func TestRunCypherSpec(t *testing.T) {
```
keremgocen, line 261: I see this is not replaced anywhere else, do we not want to verify the tool spec annotations?

internal/database/service.go
```diff
@@ -59,6 +59,47 @@ func (s *Neo4jService) ExecuteWriteQuery(ctx context.Context, cypher string, par
...
+// This allows read-only tools to determine if a query is safe to run in read-only context.
+func (s *Neo4jService) GetQueryType(ctx context.Context, cypher string, params map[string]any, database string) (string, error) {
+	if s.driver == nil {
+		err := fmt.Errorf("neo4j driver is not initialized")
+		log.Printf("error in GetQueryType: %v", err)
+		return "", err
+	}
+
+	explainedQuery := "EXPLAIN " + cypher
+	res, err := neo4j.ExecuteQuery(ctx, *s.driver, explainedQuery, params, neo4j.EagerResultTransformer, neo4j.ExecuteQueryWithDatabase(database))
```
keremgocen, line 73: Does this double the read query load? Do you think we should document that as well? Caching the query type results for identical queries could be a long-term solution but I don't think we are there yet:)

internal/database/service.go
```diff
@@ -59,6 +59,47 @@ func (s *Neo4jService) ExecuteWriteQuery(ctx context.Context, cypher string, par
...
+// GetQueryType prefixes the provided query with EXPLAIN and returns the query type (e.g. 'r' for read, 'w' for write, 'rw' etc.)
+// This allows read-only tools to determine if a query is safe to run in read-only context.
+func (s *Neo4jService) GetQueryType(ctx context.Context, cypher string, params map[string]any, database string) (string, error) {
+	if s.driver == nil {
+		err := fmt.Errorf("neo4j driver is not initialized")
+		log.Printf("error in GetQueryType: %v", err)
+		return "", err
+	}
+
+	explainedQuery := "EXPLAIN " + cypher
```
keremgocen: super nitpick: would you like to move this into a helper function and test against the formatting with the `EXPLAIN` + ` ` (empty space) + `original query` so we break tests if it's changed somehow.

internal/database/service.go
```diff
@@ -59,6 +59,47 @@ func (s *Neo4jService) ExecuteWriteQuery(ctx context.Context, cypher string, par
...
+	if res.Summary == nil {
+		err := fmt.Errorf("error during GetQueryType: no summary returned for explained query")
+		log.Printf("error during GetQueryType:: %v", err)
+		return "", err
+	}
+
+	st := res.Summary.StatementType()
+	var qType string
+	switch st {
+	case neo4j.StatementTypeReadOnly:
```
keremgocen: my final nitpick 😂: 
I'd really prefer to get these from the driver as constants instead of implementing custom strings like this, but for now, I think it's better to define them as top level constants somewhere in the file like so:
```
const (
      QueryTypeReadOnly    = "r"
      QueryTypeReadWrite   = "rw"
      QueryTypeWriteOnly   = "w"
      QueryTypeSchemaWrite = "s"
      QueryTypeUnknown     = "unknown"
  )
```
then we can use it in `read_cypher_handler.go:50` as → `if queryType != database.QueryTypeReadOnly`
which makes it easier to read and maintain imo, and same for tests.

internal/database/service.go
MacondoExpress: Counter-point(sorry I will be over critic, more as opportunity to discuss the direction, rather than on the point itself that I have zero issue in changing): 
I'm thinking that eventually it will be tested, but I thought either by:
- Unit test with mocked driver
- Integration test

I would personally found less readable if that string came from another function, but that's a  personal preference. I would be more opinionated on having separated if the goal is to unit test the string concatenation, as I personally think that would create IMO unnecessary contracts harder to modify/extend in the future for a goal that could be solved by a different testing strategy.
However, I think that be explicit: "EXPLAIN" + (space)  + `original query` would make the concatenation output more obvious rather then appending the extra space directly on the first string.

internal/database/service.go
MacondoExpress: Good shout, I think it will produce more benefits than harm adding a couple of lines of documentation, even just for the contributor's point of view!

internal/database/service.go
MacondoExpress: Good shout, I completely agree!

internal/server/server.go
MacondoExpress: I'll change it as you suggested, or we can change it entirely, I think at some point we should not document all the tools and server capabilities, we need to find the best practice here!
In the long term, we should experiment and evaluate impact on these @Liam-Doodson @mjfwebb  for visibility

internal/tools/read_cypher_handler.go
MacondoExpress: Yep

internal/database/service.go
keremgocen: not a hill I'm willing to die on...
my point was simply to detect changes otherwise we might miss 
> _...so we break tests if it's changed somehow._

I'd also prefer to test it from outside via an integration test where we call `GetQueryType` with a write query and get back an error instead (query not executed). Maybe let's leave it for now and focus that effort on integration tests?

keremgocen: what do you think about the `strings.Join([]string{"EXPLAIN", "original-query"}, delimiter)` to be a bit more explicit?
https://go.dev/play/p/Dyddvd0b9wn

MacondoExpress: > not a hill I'm willing to die on...

😂  Sorry, I didn't want to sound defensive about it. 
I think we both agree on the fact that we should test it; I think it's a good shout. The point I wanted to express is that we should not change the implementation for a temporary lack regarding the testing kit, when this will be solved more "natural" by:
- Mock the driver and add an assertion there.
- Integration test.

I'm happy to apply the proposed change. I don't particularly mind; it was merely an excuse to talk about the lack of testing tools.

TLDR of my point:
Let's build the tools first, rather than change implementation around the lack of them.

> what do you think about the strings.Join([]string{"EXPLAIN", "original-query"}, delimiter) to be a bit more explicit?

Yeah I'll do using that style!

internal/tools/run_cypher_handler_test.go
MacondoExpress: I'm not sure about this tbh, but I would suggest including that coverage within the specification files.
I'm not sure how much value it will provide, since it won't actually test functionality, however, it could help catch unintended changes. I believe we'll see more testing benefits once we begin implementing structured input and output. Happy to do add it, In that case, I would open a new PR with the spec test for schema/read and write cypher.

internal/tools/write_cypher_spec.go
MacondoExpress: > But I don't feel like we are making it clear in the README, or it could help for users to mention it even in the code as a comment where the validation happens, or in the spec if that works.

Very good point!

internal/server/server.go
Liam-Doodson: Yeah exactly, we don't need to over think this until we have some way of validating the impact of changing descriptions like this

I also think let's just go with read-cypher for now as write-cypher won't exist in read-only mode and I'd prefer it tries read-cypher before falling back to write cypher anyway

internal/database/service.go
Liam-Doodson: > Caching the query type results for identical queries could be a long-term solution but I don't think we are there yet:)

I'm pretty sure the db actually does this so I don't think it's worth us over complicating it at this point. Obviously, it still means the extra round trip but let's see how the performance is/feels with real usage and we can iterate from there

MacondoExpress: I agree on the cache mechanism for database invocations, I think it will be out of the scope for the project for long time or ever.

internal/database/service.go
```diff
@@ -59,6 +60,33 @@ func (s *Neo4jService) ExecuteWriteQuery(ctx context.Context, cypher string, par
...
+		err := fmt.Errorf("neo4j driver is not initialized")
+		log.Printf("error in GetQueryType: %v", err)
+		return neo4j.StatementTypeUnknown, err
+	}
+
+	explainedQuery := strings.Join([]string{"EXPLAIN", cypher}, " ")
+	res, err := neo4j.ExecuteQuery(ctx, *s.driver, explainedQuery, params, neo4j.EagerResultTransformer, neo4j.ExecuteQueryWithDatabase(database))
+	if err != nil {
+		wrappedErr := fmt.Errorf("error during GetQueryType: %w", err)
+		log.Printf("error during GetQueryType: %v", wrappedErr)
```
mjfwebb: Nittiest of nits: Why are the logged errors all starting with lowercase "error" instead of "Error"?

internal/tools/read_cypher_handler.go
```diff
@@ -0,0 +1,71 @@
...
+
+	// Get queryType by pre-appending "EXPLAIN" to identify if the query is of type "r", if not raise a ToolResultError
+	queryType, err := dbService.GetQueryType(ctx, Query, Params, config.Database)
+	if err != nil {
+		log.Printf("Error while classifying Cypher query: %v", err)
+		return mcp.NewToolResultError(err.Error()), nil
+	}
+
+	if queryType != neo4j.StatementTypeReadOnly { // only queryType == "r" are allowed in read-cypher
+		errMessage := "read-cypher can run only read-only Cypher statements. For write operations (CREATE, MERGE, DELETE, SET, etc...), schema/admin commands, or PROFILE queries, use write-cypher instead."
```
mjfwebb:
```suggestion
		errMessage := "read-cypher can only run read-only Cypher statements. For write operations (CREATE, MERGE, DELETE, SET, etc...), schema/admin commands, or PROFILE queries, use write-cypher instead."
```

internal/tools/read_cypher_handler.go
```diff
@@ -0,0 +1,71 @@
...
+		log.Printf("Rejected non-read query (type=%s): %s", queryType, Query)
```
mjfwebb: I think queryType here is an int

internal/tools/read_cypher_handler_test.go
```diff
@@ -0,0 +1,329 @@
...
+			},
+		}
+
+		result, err := handler(context.Background(), request)
+		if err != nil {
+			t.Errorf("Expected no error from handler, got: %v", err)
+		}
+		if result == nil || !result.IsError {
+			t.Error("Expected error result for non-read query type")
+		}
```
mjfwebb, line 298: We might want to include tests for the logged messages (log.Printf) since we utilise that to report errors to the server administrators.

internal/database/service.go
Liam-Doodson: It's a Go thing. This is the reason: https://go.dev/wiki/CodeReviewComments#error-strings

mjfwebb: I mean the `log.Printf`

keremgocen: The lowercase starting `error` (or `log`) is the package name (the short, easy-to-read named ones are usually the standard library packages), and that's how you import things from other packages. Then, you call the exported functions (usually via an interface if it's an external API and not a standard library function).
For the Neo4j driver, we use `neo4j.ExecuteQuery`, for example. Which is a function in the `neo4j` package, which should have been an interface imo, but that's a different topic:)
https://go.dev/blog/package-names

mjfwebb: That's not what I'm talking about.

I'm talking about this line of code here:

`log.Printf("error during GetQueryType: %v", wrappedErr)`

The log line is beginning with a lowercase word "error".

keremgocen: Oh, I guess that's a preference thing (and a linter setting if you want) which affects how you want your logs to appear on the console such as in GCP when you read it.
https://staticcheck.dev/docs/checks/#ST1005

and as Liam mentioned, it's recommended in Go code review guidelines, that's not mandatory, but maybe frowned upon?:)

Liam-Doodson: Oh right, yeah I think logs should be capitalised

> This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.

internal/tools/read_cypher_handler.go
MacondoExpress: nice catch

internal/tools/read_cypher_handler_test.go
MacondoExpress: This is my opinion, but I think the value that we add by testing the log message is less than the tradeoff paid in testing, especially as it will be a part in constant movement or at least is how I envision that.

PR #54, opened 2025-10-01
Add linter configuration and refactor service interfaces

keremgocen:
Introduce a simple linter configuration and address identified issues. Refactor the database service interface for consistency across the codebase.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,26 @@
+version: "2"
+run:
+  concurrency: 4
+  tests: true
+linters:
+  enable:
+    - bodyclose
+    - gocritic
+    - gosec
+    - makezero
+    - misspell
+    - nakedret
+    - revive
+  exclusions:
+    generated: lax
+    presets:
+      - comments
+      - common-false-positives
+      - legacy
+      - std-error-handling
+  settings:
+    staticcheck:
+      checks:
+        - "all"
+        - -QF1008
+        - -ST1000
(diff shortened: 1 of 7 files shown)


.golangci.yaml
```diff
@@ -0,0 +1,26 @@
+version: "2"
```
luffy1727, line 1: Why is it version 2

keremgocen: I think we are using v2 now (latest), should be the same in Aura

luffy1727: i see

luffy1727 (approved):

keremgocen:
> LGTM. But i guess you should get approval from someone who is more involved in the project. 😅

thanks for the review, it might be worth waiting for #53 to be merged as this might create some conflicts

internal/database/interfaces.go
```diff
@@ -23,8 +23,8 @@ type RecordFormatter interface {
 	Neo4jRecordsToJSON(records []*neo4j.Record) (string, error)
 }
 
-// DatabaseService combines query execution and record formatting
-type DatabaseService interface {
+// Service combines query execution and record formatting
+type Service interface {
```
MacondoExpress, line 31: Is it not too broad "Service" as a name for an interface name?

keremgocen:
> I'm questioning if "Sevice" is the right name for that interface, but apart from that it looks good to me

I think we discussed it offline but just to have visibility here, it reads better when you do `database.Service` vs `database.DatabaseService`, I think. That's why I renamed it; for now, this is the only service layer we plan to have. It was also a linter warning, just noticed it myself with the addition of golanci.yaml in this very PR:)

PR #57, opened 2025-10-09
[WIP] MCP HTTP - Auth0 flow 

keremgocen:
You'd need to manually add the MCP and the localhost URL in VSCode. The config I ended up having is below, not sure if we need all of it yet;

```
"neo4j-mcp": {
      "url": "http://127.0.0.1:8080/mcp",
      "type": "http",
      "headers": {
        "Origin": "http://127.0.0.1:50115"
      },
      "auth": {
        "type": "oauth2",
        "authorizationUrl": "http://127.0.0.1:8080/authorize",
        "tokenUrl": "http://127.0.0.1:8080/token",
        "clientId": "<from Auth0 app>",
        "scope": "openid profile",
        "redirectUri": "http://127.0.0.1:50115"
      }
    }
```

where `http://127.0.0.1:50115` is the VSCode OAuth redirect URL, which might be different for you.

And some Auth0 config is required;
```
Configure Auth0 Application (CRITICAL)
In your Auth0 Dashboard:

Go to Applications → Your App (with the clientID above)
Add to Allowed Callback URLs: http://127.0.0.1:50115
Add to Allowed Web Origins: http://127.0.0.1:50115,http://127.0.0.1:8080
Add to Allowed Origins (CORS): http://127.0.0.1:50115,http://127.0.0.1:8080
```


internal/server/oauth_proxy.go
```diff
@@ -0,0 +1,148 @@
...
+	scheme := "http"
+	if r.TLS != nil {
+		scheme = "https"
+	}
+	baseURL := scheme + "://" + r.Host
+
+	metadata := map[string]interface{}{
+		"issuer":                           "https://" + s.config.Auth0Domain + "/",
+		"authorization_endpoint":           baseURL + "/authorize",
+		"token_endpoint":                   baseURL + "/token",
```
Liam-Doodson, line 132: Should these not also be URLs on the auth server, not the client?

```suggestion

	metadata := map[string]interface{}{
		"issuer":                           "https://" + s.config.Auth0Domain + "/",
		"authorization_endpoint":           s.config.Auth0Domain + "/authorize",
		"token_endpoint":                   s.config.Auth0Domain + "/token",
```

keremgocen: Good point, and I think this currently works as you said because our OAuth proxy is already redirecting these requests (through `baseURL`) to `s.config.Auth0Domain`, which seems like an unnecessary step. Therefore, we should be able to update these redirect URLs and eliminate our own OAuth proxy. I'll look into that.

PR #61, opened 2025-10-13
[HTTP AUTH SPIKE - do not merge yet] Add HTTP transport with JWT middleware and configuration updates

keremgocen:
- adds temporary helper flags
- ResourceIdentifier in config now we try to validate it in JWT's registeredClaims (audience in auth provider)
- fixes RFC 9728, the protected resource metadata header from `Bearer realm=` to `Bearer resource_metadata=`
- adds native corsMiddleware to allow all origins, we should disable this in main I think?
- updates golangci-lint config to complain about unused stuff and todo notes 
- adds RBAC on tool calls via external auth roles and user permissions in the JWT

[Note] We are not planning to merge the extra auth flow endpoints apart from the changes in the `handleProtectedResourceMetadata`. This PR is for the rest of the team to test the auth changes locally.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/neo4j/mcp
 
-go 1.25.1
+go 1.25.3
 
 require (
 	github.com/auth0/go-jwt-middleware/v2 v2.3.0
@@ -18,8 +18,8 @@ require (
 	github.com/spf13/cast v1.10.0 // indirect
 	github.com/wk8/go-ordered-map/v2 v2.1.8 // indirect
 	github.com/yosida95/uritemplate/v3 v3.0.2 // indirect
-	golang.org/x/crypto v0.35.0 // indirect
-	golang.org/x/sync v0.16.0 // indirect
+	golang.org/x/crypto v0.43.0 // indirect
+	golang.org/x/sync v0.17.0 // indirect
 	gopkg.in/go-jose/go-jose.v2 v2.6.3 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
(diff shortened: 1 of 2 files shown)


cmd/neo4j-mcp/main.go
```diff
@@ -21,6 +21,13 @@ func main() {
 		fmt.Printf("neo4j-mcp version: %s\n", Version)
 		return
 	}
+
+	// Handle help flag
+	if len(os.Args) > 1 && (os.Args[1] == "-h" || os.Args[1] == "--help") {
```
MacondoExpress, line 26: In the case we want to merge this, I would move the flags logic outside the main file

internal/server/auth_middleware.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+	return userCtx
+}
+
+// extractPermissionsFromClaims parses permissions from validated JWT claims
+// Permissions are extracted from the scope claim as a space-separated string
+func extractPermissionsFromClaims(validatedClaims *validator.ValidatedClaims) []string {
+	customClaims, ok := validatedClaims.CustomClaims.(*CustomClaims)
+	if !ok {
+		log.Printf("⚠ Failed to extract custom claims")
```
MacondoExpress, line 44: I would remove these icons as they may create unnecessary friction when integrating with external services

internal/server/auth_middleware.go
```diff
@@ -0,0 +1,263 @@
...
+//
+// Per MCP spec and RFC 9728, the realm MUST point to this server's protected resource
+// metadata endpoint, which tells clients:
+// 1. What resource identifier to use in token requests
+// 2. Where to find the authorization server (Auth0)
+func (s *Neo4jMCPServer) sendUnauthorized(w http.ResponseWriter, errorDescription string) {
+	// Construct the protected resource metadata URL per RFC 9728
+	// This tells clients where to discover this resource server's configuration
+	// NOTE: This should point to THIS server's metadata endpoint, not Auth0's
+	metadataURL := "http://" + s.config.HTTPHost + ":" + s.config.HTTPPort + "/.well-known/oauth-protected-resource"
```
MacondoExpress, line 190: Not sure if it's worth changing or adding a FIXME/TODO comment but in the future this URL could be:
`"https://resource.example.com/.well-known/oauth-protected-resource"` - therefore with a different protocol and optional port

internal/server/auth_middleware.go
```diff
@@ -0,0 +1,263 @@
...
+
+	// Build WWW-Authenticate header value per RFC 6750 Section 3
```
MacondoExpress, line 192: Extremely nitpick, but if you think it could be helpful (I don't) the format of `WWW-Authenticate` is defined by the RFC 6750 but the `resource_metadata` value is defined by the https://datatracker.ietf.org/doc/html/rfc9728/#section-5.1

PR #71, opened 2025-10-17
Add integration tests

mjfwebb:
This PR adds integration tests and helpers to easily add new tests in the future.

Tests will use testcontainers where we spin up a Neo4j database and run our integration tests against it.

As there was a deliberate choice to use only one container (as opposed to a container for each test), there's a need to isolate tests, which we do by using unique node labels when adding data to the database. This way each test is responsible for its own data creation and cleanup.

Along with this there are various Neo4j data structure specific assertion helper functions added. This is quite opinionated, but the aim is to make it easy for contributors to add new tests.

Note that `waitForConnectibity` is a custom function which waits for Neo4j to be ready to accept connections using exponential backoff retry logic. It's used during test setup to ensure the database container is fully initialized before running tests.

### DIFF
--- a/test/integration/README.md
+++ b/test/integration/README.md
@@ -7,16 +7,19 @@ Integration tests for the Neo4j MCP server using a shared Neo4j container (inclu
 ```go
 func TestMCPIntegration_MyFeature(t *testing.T) {
     t.Parallel()
-    tc := NewTestContext(t)
+    tc := helpers.NewTestContext(t)
 
-    // Seed test data (automatically isolated and cleaned up)
-    tc.SeedNode("Person", map[string]any{"name": "Alice"})
+    // Seed test data (automatically isolated with unique labels and cleaned up)
+    personLabel, err := tc.SeedNode("Person", map[string]any{"name": "Alice"})
+    if err != nil {
+        t.Fatalf("failed to seed data: %v", err)
+    }
 
     // Call tool
-    handler := tools.ReadCypherHandler(tc.Deps)
+    handler := cypher.ReadCypherHandler(tc.Deps)
     res := tc.CallTool(handler, map[string]any{
-        "query":  "MATCH (p:Person {test_id: $testID}) RETURN p",
-        "params": map[string]any{"testID": tc.TestID},
+        "query":  "MATCH (p:" + personLabel + " {name: $name}) RETURN p",
+        "params": map[string]any{"name": "Alice"},
     })
 
     // Parse and assert
@@ -24,7 +27,8 @@ func TestMCPIntegration_MyFeature(t *testing.T) {
     tc.ParseJSONResponse(res, &records)
 
     person := records[0]["p"].(map[string]any)
-    AssertNodeProperties(t, person, map[string]any{"name": "Alice"})
+    helpers.AssertNodeProperties(t, person, map[string]any{"name": "Alice"})
+    helpers.AssertNodeHasLabel(t, person, personLabel)
 }
 ```
 
@@ -32,17 +36,18 @@ func TestMCPIntegration_MyFeature(t *testing.T) {
 
 **TestContext:**
 
-- `NewTestContext(t)` - Auto-isolation + cleanup
-- `SeedNode(label, props)` - Create test data
+- `helpers.NewTestContext(t)` - Auto-isolation + cleanup
+- `SeedNode(label, props)` - Create test data with unique label, returns `(UniqueLabel, error)`
+- `GetUniqueLabel(label)` - Get a unique label for creating nodes manually
 - `CallTool(handler, args)` - Invoke MCP tool
 - `ParseJSONResponse(res, &v)` - Parse response
 - `VerifyNodeInDB(label, props)` - Check DB state
 
 **Assertions:**
 
-- `AssertNodeProperties(t, node, props)`
-- `AssertNodeHasLabel(t, node, label)`
-- `AssertSchemaHasNodeType(t, schema, label, props)`
+- `helpers.AssertNodeProperties(t, node, props)`
+- `helpers.AssertNodeHasLabel(t, node, label)`
+- `helpers.AssertSchemaHasNodeType(t, schema, label, props)`
 
 ## Running Tests
 
@@ -75,5 +80,6 @@ go test -tags=integration ./test/integration/... -v
 ## Important
 
 - Always use `t.Parallel()` for parallel execution
-- Always include `test_id` in queries for isolation
-- Test data is automatically tagged with unique IDs and cleaned up
+- Always use the `UniqueLabel` returned by `SeedNode()` or `GetUniqueLabel()` in your queries for isolation
+- Test data is automatically tagged with unique labels and cleaned up after each test
+- Import the helpers package: `"github.com/neo4j/mcp/test/integration/helpers"`
--- a/test/integration/helpers/helpers.go
+++ b/test/integration/helpers/helpers.go
@@ -23,13 +23,23 @@ import (
 	"github.com/testcontainers/testcontainers-go/wait"
 )
 
+type UniqueLabel string
+
+// String returns the string representation of the UniqueLabel.
+// This implements the fmt.Stringer interface, making it work seamlessly with fmt functions.
+func (ul UniqueLabel) String() string {
+	return string(ul)
+}
+
 // TestContext holds common test dependencies
 type TestContext struct {
-	Ctx     context.Context
-	T       *testing.T
-	TestID  string
-	Service database.Service
-	Deps    *tools.ToolDependencies
+	Ctx           context.Context
+	T             *testing.T
+	TestID        string
+	Service       database.Service
+	Deps          *tools.ToolDependencies
+	createdLabels map[string]bool
+	labelMutex    sync.Mutex
 }
 
 var (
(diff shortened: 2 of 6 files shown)


test/integration/helpers_test.go
```diff
@@ -0,0 +1,330 @@
...
+	return fmt.Sprintf("test-%s", uuid.NewString())
+}
+
+func createNeo4jContainer(ctx context.Context) (testcontainers.Container, string, error) {
+	req := testcontainers.ContainerRequest{
+		Image:        "neo4j:5.24.2-community",
+		ExposedPorts: []string{"7687/tcp"},
+		Env: map[string]string{
+			"NEO4J_AUTH":        "neo4j/password",
+			"NEO4JLABS_PLUGINS": `["apoc","graph-data-science"]`,
```
MacondoExpress: I would accept some of these config/env as configuration argument - the immediate reason I have in mind is to test case where we want to test how we behave with missing optional requirements such as GDS.

test/integration/helpers/helpers.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	if logs != "" {
+		return fmt.Errorf("neo4j connectivity not ready: %v\ncontainer logs:\n%s", lastErr, logs)
+	}
+	return fmt.Errorf("neo4j connectivity not ready: %v", lastErr)
+}
+
+// NewTestContext creates a new test context with automatic cleanup
+func NewTestContext(t *testing.T) *TestContext {
```
MacondoExpress, line 98: Nitpick: In other project guidelines we have chosen to order methods depending to the visibility: from Public (top) to Private (bottom). Could be worth applying the same guideline even here

test/integration/helpers_test.go
```diff
@@ -0,0 +1,330 @@
...
+	"io"
+	"testing"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/neo4j/mcp/internal/database"
+	"github.com/neo4j/mcp/internal/tools"
+	"github.com/neo4j/neo4j-go-driver/v5/neo4j"
+	testcontainers "github.com/testcontainers/testcontainers-go"
```
MacondoExpress: is this alias helpful?

test/integration/helpers_test.go
```diff
@@ -0,0 +1,330 @@
...
+		_ = ctr.Terminate(ctx)
+		return nil, "", err
+	}
+
+	boltURI := fmt.Sprintf("bolt://%s:%s", host, port.Port())
+	return ctr, boltURI, nil
+}
+
+// waitForConnectivityNoT waits for Neo4j connectivity with exponential backoff.
+func waitForConnectivityNoT(ctx context.Context, ctr testcontainers.Container, drv *neo4j.DriverWithContext) error {
```
MacondoExpress: Maybe it's a stupid question, but what the `NoT` suffix stands for?

test/integration/README.md
```diff
@@ -0,0 +1,59 @@
+# Integration Tests
```
keremgocen, line 1: If we convert this file into a doc.go with the package name I think vscode will recognise the package and we don't have to change the settings.
```doc.go
// Package integration description here
package integration
```

edit: nope, that did not work

test/integration/helpers_test.go
```diff
@@ -0,0 +1,330 @@
...
+	t.Cleanup(func() {
+		tc.Cleanup()
+	})
+
+	return tc
+}
+
+// Cleanup removes all test data tagged with this test ID
+func (tc *TestContext) Cleanup() {
+	_, _ = tc.Service.ExecuteWriteQuery(
```
keremgocen: we should not ignore the cleanup errors as it might ruin other tests and hard to notice later on.

test/integration/helpers_test.go
```diff
@@ -0,0 +1,330 @@
...
+	whereClauses := []string{}
+	for key := range props {
+		whereClauses = append(whereClauses, fmt.Sprintf("n.%s = $%s", key, key))
+	}
+	whereClause := ""
+	if len(whereClauses) > 0 {
+		whereClause = " WHERE " + whereClauses[0]
+		for i := 1; i < len(whereClauses); i++ {
+			whereClause += " AND " + whereClauses[i]
+		}
```
keremgocen: we can simplify it via something like:
```suggestion
		whereClause = " WHERE " + strings.Join(whereClauses, " AND ")
```

test/integration/integration_test.go
```diff
@@ -0,0 +1,56 @@
...
+		log.Fatalf("failed to verify connectivity: %v", err)
+	}
+
+	sharedCfg = &config.Config{URI: boltURI, Username: username, Password: password, Database: "neo4j"}
+	sharedDrv = drv
+
+	code := m.Run()
+
+	_ = drv.Close(context.Background())
+	_ = ctr.Terminate(context.Background())
```
keremgocen: let's log these errors as well
```suggestion
    if err := drv.Close(context.Background()); err != nil {
      log.Printf("Warning: failed to close driver: %v", err)
    }
    if err := ctr.Terminate(context.Background()); err != nil {
      log.Printf("Warning: failed to terminate container: %v", err)
    }
```

test/integration/helpers_test.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+// makeTestID returns a unique test id suitable for tagging resources created by tests.
+func makeTestID() string {
+	return fmt.Sprintf("test-%s", uuid.NewString())
+}
+
+func createNeo4jContainer(ctx context.Context) (testcontainers.Container, string, error) {
+	req := testcontainers.ContainerRequest{
+		Image:        "neo4j:5.24.2-community",
```
keremgocen: should we get these from config as well to test against different versions?

test/integration/read_cypher_test.go
```diff
@@ -0,0 +1,36 @@
...
+	})
+
+	var records []map[string]any
+	tc.ParseJSONResponse(res, &records)
+
+	if len(records) != 1 {
+		t.Fatalf("expected 1 record, got %d", len(records))
+	}
+
+	pNode := records[0]["p"].(map[string]any)
```
keremgocen: we are doing type assertion but ignoring errors, maybe do this instead:
```suggestion
    pNode, ok := records[0]["p"].(map[string]any)
    if !ok {
        t.Fatalf("expected p to be map[string]any, got %T",
    records[0]["p"])
    }
```

keremgocen (commented):
I think we'll have to modify vscode settings.json afterall, go pls seems to be ignoring the files with the build tag `//go:build integration` without that setting as you said. I could not find a workaroud for that.

>When gopls tries to type-check this "incomplete" package, it fails because:
>The package has no real content (just documentation)
>It can't resolve types and functions referenced in the visible files
>It can't build a proper package symbol table

MacondoExpress (commented):
I personally would move the Test Helpers in a separate package/folder from the actual tests

test/integration/helpers_test.go
mjfwebb: Yeah this is inconsistent, that's on me. It denotes it's not a test helper, i.e. it doesn't invoke `t.Helper()`

Renamed to `waitForConnectivity`

test/integration/helpers_test.go
mjfwebb: Sure!

test/integration/read_cypher_test.go
mjfwebb: Honestly I deliberately decided to ignore it because it was getting so verbose haha! I still need to get used to Go

test/integration/helpers_test.go
mjfwebb: d'oh! I had this as env and then changed it back :sweat_smile: 

I'll update!

test/integration/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+The integration tests use environment variables to configure the Neo4j test container. All variables have sensible defaults:
+
+| Environment Variable            | Default                         | Description               |
+| ------------------------------- | ------------------------------- | ------------------------- |
+| `NEO4J_IMAGE`                   | `neo4j:5.24.2-community`        | Neo4j Docker image to use |
+| `NEO4J_USERNAME`                | `neo4j`                         | Database username         |
+| `NEO4J_PASSWORD`                | `password`                      | Database password         |
+| `NEO4JLABS_PLUGINS`             | `["apoc","graph-data-science"]` | Plugins to install        |
+| `NEO4J_PROCEDURES_UNRESTRICTED` | `apoc.*,gds.*`                  | Unrestricted procedures   |
```
MacondoExpress: Ignorant question, but is this needed? What happens when you don't pass it?
Same as with the NEO4J_PROCEDURES_ALLOWLIST/NEO4J_APOC_EXPORT_ENABLED etc..

MacondoExpress (changes requested):
I would personally move to a solution where the test helpers creates unique labels similar as what was done by the neo4j/graphql library.
The main reason for this asking is that by doing that we will have less side effects between tests and it will become easier to track types and dependencies.

test/integration/README.md
mjfwebb: I think we can happily remove them actually!

PR #74, opened 2025-10-21
Improve tools ergonomics

MacondoExpress:
This PR separate the cypher tools from the gds one, and delegates the server package to handle the registration logic.

### DIFF
--- a/internal/server/tool_register_test.go
+++ b/internal/server/tool_register_test.go
@@ -26,12 +26,8 @@ func TestGetAllTools(t *testing.T) {
 		s := server.NewNeo4jMCPServer("test-version", cfg, mockDB)
 
 		// Expected tools that should be registered
-		expectedTools := []string{
-			"get-schema",
-			"read-cypher",
-			"write-cypher",
-			"list-gds-procedures",
-		}
+		// update this number when a tool is added or removed.
+		expectedTotalToolsCount := 4
 
 		// Register tools
 		err := s.RegisterTools()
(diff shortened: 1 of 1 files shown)


internal/server/tool_register_test.go
```diff
@@ -0,0 +1,48 @@
...
+			"write-cypher",
+			"list-gds-procedures",
+		}
+
+		// Register tools
+		err := s.RegisterTools()
+		if err != nil {
+			t.Fatalf("RegisterTools() failed: %v", err)
+		}
+		registeredTools := len(s.MCPServer.ListTools())
```
keremgocen, line 37: nice!

internal/tools/types.go
```diff
@@ -0,0 +1,12 @@
+package tools
```
keremgocen, line 1: While this is a tool-related struct, it makes sense to move this into the server package as well to keep the tools folder clean. (I think)

internal/server/tool_register_test.go
keremgocen: While I like this check, on a second thought, if this is the only place we are using `s.MCPServer` and that's why we are making it public, I'm not sure if it's a good enough reason to expose it. We should keep it private unless it needs to be accessed from outside the package. Maybe a wrapper function that calls `s.ListServerTools()` that returns the same thing makes sense here imo.

internal/server/tools_register.go
```diff
@@ -0,0 +1,48 @@
...
+func (s *Neo4jMCPServer) RegisterTools() error {
+	deps := &tools.ToolDependencies{
+		Config:    s.config,
+		DBService: s.dbService,
+	}
+	registerAllTools(s.MCPServer, deps)
+	return nil
+}
+
+// RegisterAllTools registers all available MCP tools
```
keremgocen:
```suggestion
// registerAllTools registers all available MCP tools
```
The comment is exported, but the implementation is not

internal/tools/types.go
keremgocen: Another second thought, this might create a dependency issue as the server will import handlers from tools and tools will import ToolDependencies from the server. I think we are good for now.

internal/server/tool_register_test.go
MacondoExpress: I thought the same, but I thought that it's correct to have it public, then my reasoning were more if make it public there or behind a getter. For context, this is similar how you access the low-level MCP server in the official SDK

PR #76, opened 2025-10-24
Run integration tests in CI

mjfwebb:
This pull request updates the test workflow to distinguish between unit and integration tests, ensuring that both types are run separately during CI.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -56,7 +56,6 @@ jobs:
         uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6
         with:
           go-version: "1.25.3"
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -33,14 +32,41 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
+
+      - name: Set up Go
+        uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6
+        with:
+          go-version: "1.25.3"
+      - run: go install go.uber.org/mock/mockgen@latest
```
MacondoExpress: I think this is not needed

PR #86, opened 2025-11-03
Add custom parameters binder and improve tests

### LINKED ISSUE #70: Incorrect type coercion for Cypher query parameters
**Description:**

When using the `read-cypher` an other Cypher tools integer parameters for the Cypher query are being incorrectly coerced into floating-point numbers. This causes queries that expect integer arguments (like `LIMIT`) to fail.

**Steps to Reproduce:**

1.  Send a `tools/call` request to the `read-cypher` tool.
2.  Provide a Cypher query that uses an integer parameter, for example `MATCH(n) RETURN n LIMIT $limit`.
3.  Provide the parameter as an integer in the `params` object, e.g., `"params": { "limit": 1 }`.

**JSON-RPC Request/Response:**

_Request_:

```json
{
  "jsonrpc": "2.0",
  "method": "tools/call",
  "id": 5,
  "params": {
    "name": "read-cypher",
    "arguments": {
      "query": "MATCH(n) RETURN n LIMIT $limit",
      "params": { "limit": 1 }
    }
  }
}
```

_Response_:

```json
{
  "jsonrpc": "2.0",
  "id": 5,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "failed to execute read query: Neo4jError: Neo.ClientError.Statement.ArgumentError (Invalid input. '1.0' is not a valid value. Must be a non-negative integer.)"
      }
    ],
    "isError": true
  }
}
```

**Expected Behavior:**

The query should be executed successfully, with the `limit` parameter being treated as an integer (`1`).

**Actual Behavior:**

The query fails with a `Neo.ClientError.Statement.ArgumentError`. The error message "Invalid input. '1.0' is not a valid value. Must be a non-negative integer." indicates that the integer `1` was converted to a float `1.0` before being passed to the Neo4j driver.

**MCP Version:**

0.3.0

**OS/Arc:**

MacOS/ARM

keremgocen:
Fixed integer parameter conversion issue where `{"limit": 1}` was incorrectly converted to `float64(1.0)` instead of `int64(1)`, causing Neo4j "Invalid input. '1.0' is not a valid value" errors.

- introduced a custom parameters binder that correctly parses integers using `json.Number`.
- fixed the skipped integration test and added unit test coverage for the `BindArguments` function. (thanks copilot)

fixes #70

root cause: golang's json.v2 is converting numbers to float64 by default. Alternatively, we could have used https://pkg.go.dev/encoding/json/v2#WithMarshalers or a custom unmarshaller, but both approaches require not using mark3labs's `request.BindArguments()`.

>unmarshal uses a float64 by default to represent a JSON number. To change the default, specify the [WithUnmarshalers](https://pkg.go.dev/encoding/json/v2#WithUnmarshalers) option with a custom unmarshaler that pre-populates the interface value with a concrete Go type that can preserve precision.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/mark3labs/mcp-go v0.42.0
 	github.com/neo4j/neo4j-go-driver/v5 v5.28.4
+	github.com/stretchr/testify v1.11.1
 	github.com/testcontainers/testcontainers-go v0.39.0
 	go.uber.org/mock v0.6.0
 )
@@ -53,7 +54,6 @@ require (
 	github.com/shirou/gopsutil/v4 v4.25.9 // indirect
 	github.com/sirupsen/logrus v1.9.3 // indirect
 	github.com/spf13/cast v1.10.0 // indirect
-	github.com/stretchr/testify v1.11.1 // indirect
 	github.com/tklauser/go-sysconf v0.3.15 // indirect
 	github.com/tklauser/numcpus v0.10.0 // indirect
 	github.com/wk8/go-ordered-map/v2 v2.1.8 // indirect
--- a/internal/tools/cypher/params_converter.go
+++ b/internal/tools/cypher/params_converter.go
@@ -3,6 +3,7 @@ package cypher
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 
 	"github.com/mark3labs/mcp-go/mcp"
 )
@@ -33,8 +34,12 @@ func BindArguments(request mcp.CallToolRequest, target any) error {
 	if converter, ok := target.(ParamsConverter); ok {
 		params := converter.GetParams()
 		if params != nil {
-			convertedParams := ConvertNumbers(params).(map[string]any)
-			converter.SetParams(convertedParams)
+			if convertedParams, ok := ConvertNumbers(params).(map[string]any); ok {
+				converter.SetParams(convertedParams)
+			} else {
+				// Optionally handle the error case, e.g., log or return an error
+				return fmt.Errorf("failed to convert params to map[string]any")
+			}
 		}
 	}
 


test/integration/issue_70_test.go
```diff
@@ -27,7 +27,7 @@ func TestIssue70(t *testing.T) {
 		},
 		{
 			name:    "write-cypher",
-			handler: cypher.ReadCypherHandler,
+			handler: cypher.WriteCypherHandler,
```
MacondoExpress, line 30: ah good catch, that was slipped!

MacondoExpress (approved):

test/integration/issue_70_test.go
keremgocen: Do we want to keep tests for specific bugs in their own files like this (`issue_80_test.go`) or do we want to keep them in their corresponding test file instead? (this also depends on how many bugs we will have but I'm questioning the idea of having a dedicated test file per bug -- assuming there are many bugs eventually)

mjfwebb: I have no strong opinion on this, but here's some things to consider:

One of the benefits of creating a test file specific to an issue is that it can be otherwise difficult to identify and name the circumstances of a bug, so it allows you to just throw in a test without having to concern yourself with naming and placing the test correctly. Naming it the same as an issue also allows one person to create a failing test and someone else to easily gather context and work on it later.

On the other hand, if we have a nice structure and we find it easy to add to our standard suite of tests, I'm all for that. We could cover the context issue by adding helpful test descriptions and comments.

keremgocen: merging the PR, as I don't think this discussion was blocking it, we can discuss this further if it starts becoming an issue to do individual test files. One immediate change we could do is to make sure test files are organised under their corresponding folders (e.g. `get_schema_test.go` should go under `test/integration/tools/get_schema_test.go`) but that still belongs to a different PR I believe.

PR #87, opened 2025-11-10
Add CLI argument handling and help command

mjfwebb:
First pass of a help command, loosely based on `go help`

### DIFF
--- a/.changes/unreleased/Patch-20251111-112806.yaml
+++ b/.changes/unreleased/Patch-20251111-112806.yaml
@@ -1,3 +1,3 @@
-kind: Minor
+kind: Patch
 body: Adds a help command available using -h or --help.
-time: 2025-11-11T10:38:11.94598821+01:00
+time: 2025-11-11T11:28:06.19896141+01:00


internal/cli/args_test.go
```diff
@@ -0,0 +1,78 @@
...
+		},
+		{
+			name:     "help flag short form",
+			args:     []string{"neo4j-mcp", "-h"},
+			version:  "1.0.0",
+			expected: true,
+		},
+		{
+			name:     "help flag long form",
+			args:     []string{"neo4j-mcp", "--help"},
```
keremgocen: we probably want to test for the help text here instead of the version?
you can check coverage via `go test -coverprofile cover.out` or right click somewhere inside the file in vscode and click on `toggle unit test coverage for this file` or something like that

internal/cli/args_test.go
```diff
@@ -0,0 +1,78 @@
...
+		},
+		{
+			name:     "unknown flag",
+			args:     []string{"neo4j-mcp", "-x"},
+			version:  "1.0.0",
+			expected: false,
+		},
+		{
+			name:     "multiple arguments with flag",
+			args:     []string{"neo4j-mcp", "-v", "extra"},
```
keremgocen: do we want `-v extra` to work?

internal/cli/args_test.go
mjfwebb: Ah thanks, I'll do this.

internal/cli/args_test.go
mjfwebb: We can make it stricter. At the moment it's very permissive and won't break if users accidentally add extra arguments.

I went with being permissive since we don't actually take any args other than flags right now.

internal/cli/args.go
```diff
@@ -5,6 +5,9 @@ import (
 	"os"
 )
 
+// osExit is a variable that can be mocked in tests
+var osExit = os.Exit
```
keremgocen, line 9: putting os.Exit into a variable felt a bit weird at first glance but this is for mocking in tests I guess?

mjfwebb: Yes as the comment above it says :smile: 

I can remove it, but we will not get test coverage then. I honestly would prefer not to put it into a variable, but it was specifically for testing.

keremgocen: to answer my own comment:
> this is actually a standard testing pattern in Go for testing functions that call os.Exit

so maybe I did not test os.Exit enough then:)

keremgocen (approved):
LGTM

keremgocen (approved):

PR #92, opened 2025-11-11
Add opt-out telemetry for product improvement

MacondoExpress:
This pull request introduces an opt-out telemetry feature to neo4j-mcp. By collecting anonymous usage data, we can better understand how the tool is used and prioritize future improvements.

Telemetry Collection: A new analytics service is added to send anonymous events to Mixpanel.

**Data Collected:** We collect non-sensitive, anonymous data, including:
Tool usage (e.g., read-cypher, get-schema).
Server startup events.
Operating System and CPU architecture.
Whether the connection is to an Aura instance.
Opt-Out by Default: Telemetry is enabled by default but can be easily disabled.
Configuration: To disable telemetry, users can set the NEO4J_TELEMETRY environment variable to "false".

Build Process: The Goreleaser configuration is updated to inject the Mixpanel token and endpoint during the official build process

### DIFF
--- a/internal/analytics/analytics_test.go
+++ b/internal/analytics/analytics_test.go
@@ -9,14 +9,14 @@ import (
 	"testing"
 
 	"github.com/neo4j/mcp/internal/analytics"
-	analytics_mocks "github.com/neo4j/mcp/internal/analytics/mocks"
+	amocks "github.com/neo4j/mcp/internal/analytics/mocks"
 	"go.uber.org/mock/gomock"
 )
 
 func TestAnalytics(t *testing.T) {
 	t.Run("EmitEvent should not send event if disabled", func(t *testing.T) {
 		ctrl := gomock.NewController(t)
-		mockClient := analytics_mocks.NewMockHTTPClient(ctrl)
+		mockClient := amocks.NewMockHTTPClient(ctrl)
 
 		analyticsService := analytics.NewAnalyticsWithClient("test-token", "http://localhost", mockClient, false)
 		analyticsService.Disable()
@@ -25,7 +25,7 @@ func TestAnalytics(t *testing.T) {
 
 	t.Run("EmitEvent should send event if enabled", func(t *testing.T) {
 		ctrl := gomock.NewController(t)
-		mockClient := analytics_mocks.NewMockHTTPClient(ctrl)
+		mockClient := amocks.NewMockHTTPClient(ctrl)
 
 		mockClient.EXPECT().Post(gomock.Any(), gomock.Any(), gomock.Any()).Return(&http.Response{
 			StatusCode: http.StatusOK,
@@ -38,7 +38,7 @@ func TestAnalytics(t *testing.T) {
 
 	t.Run("EmitEvent should send the correct event in the body", func(t *testing.T) {
 		ctrl := gomock.NewController(t)
-		mockClient := analytics_mocks.NewMockHTTPClient(ctrl)
+		mockClient := amocks.NewMockHTTPClient(ctrl)
 
 		event := analytics.TrackEvent{
 			Event: "specific_event",
@@ -87,7 +87,7 @@ func TestAnalytics(t *testing.T) {
 
 	t.Run("EmitEvent should send the correct event in the body", func(t *testing.T) {
 		ctrl := gomock.NewController(t)
-		mockClient := analytics_mocks.NewMockHTTPClient(ctrl)
+		mockClient := amocks.NewMockHTTPClient(ctrl)
 
 		event := analytics.TrackEvent{
 			Event: "specific_event",
@@ -160,7 +160,7 @@ func TestAnalytics(t *testing.T) {
 		for _, tc := range testCases {
 			t.Run(tc.name, func(t *testing.T) {
 				ctrl := gomock.NewController(t)
-				mockClient := analytics_mocks.NewMockHTTPClient(ctrl)
+				mockClient := amocks.NewMockHTTPClient(ctrl)
 
 				mockClient.EXPECT().Post(tc.expectedURL, gomock.Any(), gomock.Any()).Return(&http.Response{
 					StatusCode: http.StatusOK,
--- a/internal/database/service_test.go
+++ b/internal/database/service_test.go
@@ -6,7 +6,7 @@ import (
 	"testing"
 
 	"github.com/neo4j/mcp/internal/database"
-	database_mocks "github.com/neo4j/mcp/internal/database/mocks"
+	db "github.com/neo4j/mcp/internal/database/mocks"
 	"github.com/neo4j/neo4j-go-driver/v5/neo4j"
 	"go.uber.org/mock/gomock"
 )
@@ -18,7 +18,7 @@ func TestDatabaseService_ExecuteReadQuery(t *testing.T) {
 	ctx := context.Background()
 
 	t.Run("successful read query execution", func(t *testing.T) {
-		mockService := database_mocks.NewMockService(ctrl)
+		mockService := db.NewMockService(ctrl)
 		expectedRecords := []*neo4j.Record{}
 
 		mockService.EXPECT().
@@ -36,7 +36,7 @@ func TestDatabaseService_ExecuteReadQuery(t *testing.T) {
 	})
 
 	t.Run("query execution error", func(t *testing.T) {
-		mockService := database_mocks.NewMockService(ctrl)
+		mockService := db.NewMockService(ctrl)
 
 		mockService.EXPECT().
 			ExecuteReadQuery(ctx, "MATCH (n) RETURN n", nil).
@@ -50,7 +50,7 @@ func TestDatabaseService_ExecuteReadQuery(t *testing.T) {
 	})
 
 	t.Run("query with parameters - find person by name", func(t *testing.T) {
-		mockService := database_mocks.NewMockService(ctrl)
+		mockService := db.NewMockService(ctrl)
 
 		mockRecords := []*neo4j.Record{
 			{
@@ -89,7 +89,7 @@ func TestDatabaseService_ExecuteReadQuery(t *testing.T) {
 	})
 
 	t.Run("cypher syntax error", func(t *testing.T) {
-		mockService := database_mocks.NewMockService(ctrl)
+		mockService := db.NewMockService(ctrl)
 
 		mockService.EXPECT().
 			ExecuteReadQuery(
@@ -134,7 +134,7 @@ func TestDatabaseService_ExecuteWriteQuery(t *testing.T) {
 	ctx := context.Background()
 
 	t.Run("successful write query execution", func(t *testing.T) {
-		mockService := database_mocks.NewMockService(ctrl)
+		mockService := db.NewMockService(ctrl)
 		expectedRecords := []*neo4j.Record{}
 
 		mockService.EXPECT().
@@ -152,7 +152,7 @@ func TestDatabaseService_ExecuteWriteQuery(t *testing.T) {
 	})
 
 	t.Run("query execution error", func(t *testing.T) {
-		mockService := database_mocks.NewMockService(ctrl)
+		mockService := db.NewMockService(ctrl)
 
 		mockService.EXPECT().
 			ExecuteWriteQuery(ctx, "CREATE (n:Test)", nil).
@@ -166,7 +166,7 @@ func TestDatabaseService_ExecuteWriteQuery(t *testing.T) {
 	})
 
 	t.Run("create node with properties and return it", func(t *testing.T) {
-		mockService := database_mocks.NewMockService(ctrl)
+		mockService := db.NewMockService(ctrl)
 
 		mockRecords := []*neo4j.Record{
 			{
@@ -205,7 +205,7 @@ func TestDatabaseService_ExecuteWriteQuery(t *testing.T) {
 	})
 
 	t.Run("cypher syntax error", func(t *testing.T) {
-		mockService := database_mocks.NewMockService(ctrl)
+		mockService := db.NewMockService(ctrl)
 
 		mockService.EXPECT().
 			ExecuteWriteQuery(
--- a/internal/server/server_test.go
+++ b/internal/server/server_test.go
@@ -3,7 +3,7 @@ package server_test
 import (
 	"testing"
 
-	analytics_mocks "github.com/neo4j/mcp/internal/analytics/mocks"
+	analytics "github.com/neo4j/mcp/internal/analytics/mocks"
 	"github.com/neo4j/mcp/internal/config"
 	db_mock "github.com/neo4j/mcp/internal/database/mocks"
 	"github.com/neo4j/mcp/internal/server"
@@ -22,7 +22,7 @@ func TestNewNeo4jMCPServer(t *testing.T) {
 	}
 
 	mockDB := db_mock.NewMockService(ctrl)
-	analyticsService := analytics_mocks.NewMockService(ctrl)
+	analyticsService := analytics.NewMockService(ctrl)
 	analyticsService.EXPECT().EmitEvent(gomock.Any()).AnyTimes()
 	analyticsService.EXPECT().NewStartupEvent().AnyTimes()
 	t.Run("creates server successfully", func(t *testing.T) {
@@ -98,7 +98,7 @@ func TestNewNeo4jMCPServerEvents(t *testing.T) {
 	}
 
 	mockDB := db_mock.NewMockService(ctrl)
-	analyticsService := analytics_mocks.NewMockService(ctrl)
+	analyticsService := analytics.NewMockService(ctrl)
 
 	t.Run("emits startup and OSInfoEvent and StartupEvent events on start", func(t *testing.T) {
 		analyticsService.EXPECT().NewStartupEvent().Times(1)
--- a/internal/tools/cypher/get_schema_handler_test.go
+++ b/internal/tools/cypher/get_schema_handler_test.go
@@ -6,8 +6,8 @@ import (
 	"testing"
 
 	"github.com/mark3labs/mcp-go/mcp"
-	analytics_mocks "github.com/neo4j/mcp/internal/analytics/mocks"
-	database_mocks "github.com/neo4j/mcp/internal/database/mocks"
+	analytics "github.com/neo4j/mcp/internal/analytics/mocks"
+	db "github.com/neo4j/mcp/internal/database/mocks"
 
 	"github.com/neo4j/mcp/internal/tools"
 	"github.com/neo4j/mcp/internal/tools/cypher"
@@ -17,13 +17,13 @@ import (
 
 func TestGetSchemaHandler(t *testing.T) {
 	ctrl := gomock.NewController(t)
-	analyticsService := analytics_mocks.NewMockService(ctrl)
+	analyticsService := analytics.NewMockService(ctrl)
 	analyticsService.EXPECT().NewToolsEvent("get-schema").AnyTimes()
 	analyticsService.EXPECT().EmitEvent(gomock.Any()).AnyTimes()
 	defer ctrl.Finish()
 
 	t.Run("successful schema retrieval", func(t *testing.T) {
-		mockDB := database_mocks.NewMockService(ctrl)
+		mockDB := db.NewMockService(ctrl)
 		mockDB.EXPECT().
 			ExecuteReadQuery(gomock.Any(), gomock.Any(), gomock.Nil()).
 			Return([]*neo4j.Record{
@@ -53,7 +53,7 @@ func TestGetSchemaHandler(t *testing.T) {
 	})
 
 	t.Run("database query failure", func(t *testing.T) {
-		mockDB := database_mocks.NewMockService(ctrl)
+		mockDB := db.NewMockService(ctrl)
 		mockDB.EXPECT().
 			ExecuteReadQuery(gomock.Any(), gomock.Any(), gomock.Nil()).
 			Return(nil, errors.New("connection failed"))
@@ -75,7 +75,7 @@ func TestGetSchemaHandler(t *testing.T) {
 	})
 
 	t.Run("JSON formatting failure", func(t *testing.T) {
-		mockDB := database_mocks.NewMockService(ctrl)
+		mockDB := db.NewMockService(ctrl)
 		mockDB.EXPECT().
 			ExecuteReadQuery(gomock.Any(), gomock.Any(), gomock.Nil()).
 			Return([]*neo4j.Record{
@@ -121,7 +121,7 @@ func TestGetSchemaHandler(t *testing.T) {
 		}
 	})
 	t.Run("nil analytics service", func(t *testing.T) {
-		mockDB := database_mocks.NewMockService(ctrl)
+		mockDB := db.NewMockService(ctrl)
 		deps := &tools.ToolDependencies{
 			DBService:        mockDB,
 			AnalyticsService: nil,
@@ -138,10 +138,10 @@ func TestGetSchemaHandler(t *testing.T) {
 		}
 	})
 	t.Run("No records returned from apoc query (empty database)", func(t *testing.T) {
-		analyticsService := analytics_mocks.NewMockService(ctrl)
+		analyticsService := analytics.NewMockService(ctrl)
 		analyticsService.EXPECT().NewToolsEvent("get-schema").Times(1)
 		analyticsService.EXPECT().EmitEvent(gomock.Any()).Times(1)
-		mockDB := database_mocks.NewMockService(ctrl)
+		mockDB := db.NewMockService(ctrl)
 		mockDB.EXPECT().
 			ExecuteReadQuery(gomock.Any(), gomock.Any(), gomock.Nil()).
 			Return([]*neo4j.Record{}, nil)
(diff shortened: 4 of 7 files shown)


internal/analytics/analytics.go
```diff
@@ -0,0 +1,119 @@
...
+		return
+	}
+	trackEvents := []TrackEvent{
+		event,
+	}
+
+	log.Printf("Sending %s event to Neo4j", event.Event)
+	err := a.sendTrackEvent(trackEvents)
+	if err != nil {
+		sendErr := fmt.Errorf("error while sending analytics events for analytics purpose: %s", err.Error())
```
mjfwebb:
```suggestion
		sendErr := fmt.Errorf("error while sending analytics events for analytics: %s", err.Error())
```

internal/analytics/analytics.go
```diff
@@ -0,0 +1,119 @@
...
+
+func (a *analytics) Disable() {
+	a.disabled = true
+}
+
+// Eventually we can use mixpanel SDK
+func (a *analytics) sendTrackEvent(events []TrackEvent) error {
+	b, err := json.Marshal(events)
+	if err != nil {
+		return fmt.Errorf("error appear while marshalling track event: %w", err)
```
mjfwebb:
```suggestion
		return fmt.Errorf("error while marshalling track event: %w", err)
```

internal/analytics/analytics.go
```diff
@@ -0,0 +1,119 @@
...
+	_ = json.Unmarshal(bodyBytes, &data)
+
+	log.Printf("Response from Neo4j, Status: %s, Body: %s, Data: %d", resp.Status, string(bodyBytes), data)
+	return nil
+}
+
+func getDistinctID() string {
+	distinctID, err := uuid.NewV6()
+	if err != nil {
+		log.Printf("error while generating distinct id for analytics purpose: %s", err.Error())
```
mjfwebb:
```suggestion
		log.Printf("error while generating distinct id for analytics: %s", err.Error())
```

internal/analytics/analytics_test.go
```diff
@@ -0,0 +1,195 @@
...
+		if props["$os"] != runtime.GOOS {
+			t.Errorf("unexpected os: got %v, want %v", props["os"], runtime.GOOS)
+		}
+		if props["os_arch"] != runtime.GOARCH {
+			t.Errorf("unexpected os_arch: got %v, want %v", props["os_arch"], runtime.GOARCH)
+		}
+		if props["isAura"] == false {
+			t.Errorf("unexpected aura: got %v, want %v", props["isAura"], true)
+		}
+		assertBaseProperties(t, event.Properties)
```
mjfwebb: `assertBaseProperties` is called twice in this `Run`?

internal/analytics/analytics_test.go
```diff
@@ -0,0 +1,195 @@
...
+		if props["$os"] != runtime.GOOS {
+			t.Errorf("unexpected os: got %v, want %v", props["os"], runtime.GOOS)
+		}
+		if props["os_arch"] != runtime.GOARCH {
+			t.Errorf("unexpected os_arch: got %v, want %v", props["os_arch"], runtime.GOARCH)
+		}
+		if props["isAura"] == true {
+			t.Errorf("unexpected aura: got %v, want %v", props["isAura"], false)
+		}
+		assertBaseProperties(t, event.Properties)
```
mjfwebb: `assertBaseProperties` is called twice in this `Run`?

internal/analytics/events.go
```diff
@@ -0,0 +1,93 @@
...
+		OS:         runtime.GOOS,
+		OSArch:     runtime.GOARCH,
+		IsAura:     cfg.isAura,
+	}
+}
+
+func newInsertID() string {
+	insertID, err := uuid.NewV6()
+	if err != nil {
+		insertIDerr := fmt.Errorf("error while generating uuid analytics events for analytics purpose: %s", err.Error())
```
mjfwebb:
```suggestion
		insertIDerr := fmt.Errorf("error while generating uuid analytics events for analytics: %s", err.Error())
```

internal/analytics/interfaces.go
```diff
@@ -0,0 +1,23 @@
...
+	Disable()
+	Enable()
+	EmitEvent(event TrackEvent)
+	NewGDSProjCreatedEvent() TrackEvent
+	NewGDSProjDropEvent() TrackEvent
+	NewStartupEvent() TrackEvent
+	NewToolsEvent(toolsUsed string) TrackEvent
+}
+
+// dummy http client interface for our testing purpose
```
mjfwebb:
```suggestion
// dummy http client interface for our testing purposes
```

internal/analytics/events.go
```diff
@@ -0,0 +1,93 @@
...
+	"strings"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+const eventNamePrefix = "MCP4NEO4J"
+
+// baseProperties are the base properties attached to a MixPanel "track" event.
+// DistinctID it's a distinct ID used to identify unique users, we do not use this information, therefore for us it will distinct different executions
```
mjfwebb:
```suggestion
// DistinctID is a distinct ID used to identify unique users, we do not use this information, therefore for us it will be distinct different executions.
```

internal/analytics/events.go
```diff
@@ -0,0 +1,93 @@
...
+// InsertID it's used to deduplicate duplicate messages.
```
mjfwebb:
```suggestion
// InsertID is used to deduplicate duplicate messages.
```

internal/analytics/analytics.go
```diff
@@ -0,0 +1,119 @@
+package analytics
+
+// Package analytics abstracts analytics handling for the repository.
```
mjfwebb:
```suggestion
// Package analytics abstracts analytics handling for the program.
```

cmd/neo4j-mcp/main.go
```diff
@@ -49,9 +54,20 @@ func main() {
...
 	}
+	isAura := strings.Contains(cfg.URI, "database.neo4j.io")
+	anService := analytics.NewAnalytics(MixPanelToken, MixPanelEndpoint, isAura)
+
+	if cfg.Telemetry == "false" || MixPanelEndpoint == "" || MixPanelToken == "" {
+		log.Println("Telemetry disabled.")
+		anService.Disable()
+	} else if cfg.Telemetry == "true" {
+		anService.Enable()
+		log.Println("Telemetry is enabled to help us improve the product by collecting anonymous usage data such as: tools in use, OS and CPU architecture.")
```
mjfwebb:
```suggestion
		log.Println("Telemetry is enabled to help us improve the product by collecting anonymous usage data such as: tools being used, the operating system, and CPU architecture.")
```

Copying the wording from the README so it's consistent :)

internal/analytics/analytics.go
```diff
@@ -0,0 +1,119 @@
...
+}
+func (a *analytics) Enable() {
+	a.disabled = false
+}
+
+func (a *analytics) Disable() {
+	a.disabled = true
+}
+
+// Eventually we can use mixpanel SDK
```
mjfwebb: Instead of this comment here, instead let's open a task

internal/tools/cypher/read_cypher_handler.go
```diff
@@ -3,20 +3,34 @@ package cypher
...
 }
 
-func handleReadCypher(ctx context.Context, request mcp.CallToolRequest, dbService database.Service) (*mcp.CallToolResult, error) {
+func handleReadCypher(ctx context.Context, request mcp.CallToolRequest, dbService database.Service, asService analytics.Service) (*mcp.CallToolResult, error) {
+	if asService == nil {
+		errMessage := "Analytics service is not initialized"
+		log.Printf("%s", errMessage)
+		return mcp.NewToolResultError(errMessage), nil
+	}
+	if dbService == nil {
```
mjfwebb, line 27: Nice moving this up 👍

cmd/neo4j-mcp/main.go
```diff
@@ -49,9 +54,20 @@ func main() {
 		log.Printf("Failed to create database service: %v", err)
 		return
 	}
+	isAura := strings.Contains(cfg.URI, "database.neo4j.io")
+	anService := analytics.NewAnalytics(MixPanelToken, MixPanelEndpoint, isAura)
+
+	if cfg.Telemetry == "false" || MixPanelEndpoint == "" || MixPanelToken == "" {
+		log.Println("Telemetry disabled.")
+		anService.Disable()
+	} else if cfg.Telemetry == "true" {
```
mjfwebb, line 63: What happens if cfg.Telemetry is "potato"?

mjfwebb (commented):
Very nice! No real comments about the code structure itself.

internal/analytics/analytics.go
MacondoExpress: Maybe I phrased the comment poorly, but I meant that is also fine to use the SDK if we want, I personally don't consider an improvement as I found it less documented and less testable.

cmd/neo4j-mcp/main.go
MacondoExpress: good point I think we should validate the environment variables in our config component so that we can coerce them to the explicit type (bool). I'll add a card for that and add an explicit check in the meantime here!

MacondoExpress: https://linear.app/neo4j/issue/DEVSURF-873/support-different-types-in-the-mcp-config

internal/analytics/analytics.go
mjfwebb: If you don't think we should use it, then let's just remove the comment :)

.golangci.yaml
```diff
@@ -19,6 +19,10 @@ linters:
       - legacy
       - std-error-handling
   settings:
+    revive:
+      rules:
+        - name: unexported-return
```
keremgocen: I'm not sure about this, why do we want it?

internal/analytics/analytics.go
```diff
@@ -0,0 +1,119 @@
...
+// for testing purposes - enables dependency injection of http client
+func NewAnalyticsWithClient(mixPanelToken string, mixpanelEndpoint string, client HTTPClient, isAura bool) Service {
+	distinctID := getDistinctID()
+	cfg := analyticsConfig{
+		token:            mixPanelToken,
+		mixpanelEndpoint: mixpanelEndpoint,
+		distinctID:       distinctID,
+		startupTime:      time.Now().Unix(),
+		client:           client,
+		isAura:           isAura,
```
keremgocen, line 42: is this currently used? if it's for Aura should we wait until Aura changes to see if this is definitely needed? My reasoning is it's harder to remove things later than not adding it in the first place.

internal/analytics/analytics.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	trackEvents := []TrackEvent{
+		event,
+	}
+
+	log.Printf("Sending %s event to Neo4j", event.Event)
+	err := a.sendTrackEvent(trackEvents)
+	if err != nil {
+		sendErr := fmt.Errorf("error while sending analytics events for analytics: %s", err.Error())
+		log.Printf("analytics error: %s", sendErr.Error())
```
keremgocen, line 74: I was wondering if we want to return the error from here, but I guess it would make things even worse unless we want to stop execution at an error event and for each event you'd have to do `if err != nil`:)

We might have to pass the logger to the analytics service and log these from there too, and that's enough for now. Can't think of an analytics event we want to exit from in an error.

internal/analytics/analytics.go
```diff
@@ -0,0 +1,119 @@
...
+func (a *analytics) Enable() {
+	a.disabled = false
+}
+
+func (a *analytics) Disable() {
+	a.disabled = true
+}
+
+// Eventually we can use mixpanel SDK
+func (a *analytics) sendTrackEvent(events []TrackEvent) error {
```
keremgocen: nit: function comments should start with the function name `// sendTrackEvent` as part of the default linter I think

internal/analytics/analytics.go
```diff
@@ -0,0 +1,119 @@
...
+	b, err := json.Marshal(events)
+	if err != nil {
+		return fmt.Errorf("error while marshalling track event: %w", err)
+	}
+	url := strings.TrimRight(a.cfg.mixpanelEndpoint, "/") + "/track"
```
keremgocen, line 90: I'd like to cover this bit in a unit test if we are not already

internal/analytics/analytics.go
```diff
@@ -0,0 +1,119 @@
...
+	defer resp.Body.Close()
+
+	bodyBytes, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("read response body: %w", err)
+	}
+
+	// try to decode numeric response, fallback to raw body logging
+	var data int32
+	_ = json.Unmarshal(bodyBytes, &data)
```
keremgocen: we should not ignore these errors and handle them when we can imo, even if it's not very critical
at least log the error if nothing else

internal/analytics/analytics.go
```diff
@@ -0,0 +1,119 @@
...
+
+	log.Printf("Response from Neo4j, Status: %s, Body: %s, Data: %d", resp.Status, string(bodyBytes), data)
+	return nil
+}
+
+func getDistinctID() string {
```
keremgocen, line 114: sounds like a good place to return an error this one now, so the caller can handle it.

internal/analytics/analytics_test.go
```diff
@@ -0,0 +1,193 @@
...
+		analyticsService.EmitEvent(event)
+	})
+}
+
+func TestEventCreation(t *testing.T) {
+	analyticsService := analytics.NewAnalyticsWithClient("test-token", "http://localhost", nil, false)
+
+	t.Run("NewGDSProjCreatedEvent", func(t *testing.T) {
+		event := analyticsService.NewGDSProjCreatedEvent()
+		if event.Event != "MCP4NEO4J_GDS_PROJ_CREATED" {
```
keremgocen, line 182: where are we defining these event names?

internal/analytics/events.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+type TrackEvent struct {
+	Event      string      `json:"event"`
+	Properties interface{} `json:"properties"`
+}
+
+func (a *analytics) NewGDSProjCreatedEvent() TrackEvent {
+	return TrackEvent{
+		Event:      strings.Join([]string{eventNamePrefix, "GDS_PROJ_CREATED"}, "_"),
```
keremgocen, line 40: I think I'd prefer a central place where these constants are defined and pulled from it would be easier to read and maintain imo

internal/analytics/events.go
```diff
@@ -0,0 +1,93 @@
...
+		OSArch:     runtime.GOARCH,
+		IsAura:     cfg.isAura,
+	}
+}
+
+func newInsertID() string {
+	insertID, err := uuid.NewV6()
+	if err != nil {
+		insertIDerr := fmt.Errorf("error while generating uuid analytics events for analytics: %s", err.Error())
+		log.Printf("MixPanel error: %s", insertIDerr.Error())
```
keremgocen: also nit: I feel like we are not using the error other than the message; we could maybe just log the message in `err.Error()`, do we need the additional context?

internal/config/config.go
```diff
@@ -7,18 +7,23 @@ import (
...
+	Telemetry string // if false, disables telemetry
 }
 
 // Validate validates the configuration and returns an error if invalid
 func (c *Config) Validate() error {
 	if c == nil {
 		return fmt.Errorf("configuration is required but was nil")
 	}
+	// TODO: support different Config types with validation https://linear.app/neo4j/issue/DEVSURF-873/support-different-types-in-the-mcp-config
+	if c.Telemetry != "false" && c.Telemetry != "true" {
```
keremgocen, line 24: Why not convert `c.Telemetry` into a `bool`?

(then we also need a boolean version of the `GetEnvWithDefault` or a generic one which checks the type as well)

internal/analytics/analytics.go
MacondoExpress: Yes it is used, 
this info is to check if the user uses an Aura database or not, from @LackOfMorals requirements,
not sure if @LackOfMorals prefer a different name for the property however

internal/analytics/interfaces.go
```diff
@@ -0,0 +1,23 @@
+package analytics
+
+//go:generate mockgen -destination=mocks/mock_analytics.go -package=analytics_mocks -typed github.com/neo4j/mcp/internal/analytics Service,HTTPClient
+import (
+	"io"
+	"net/http"
+)
+
+// Service
+type Service interface {
```
keremgocen, line 10: I'd call this `Analytics` as it's the analytics interface instead

internal/analytics/analytics.go
```diff
@@ -0,0 +1,119 @@
...
+type analyticsConfig struct {
+	token            string
+	mixpanelEndpoint string
+	distinctID       string
+	startupTime      int64
+	client           HTTPClient
+	isAura           bool
+}
+
+type analytics struct {
```
keremgocen: Then we can call this `AnalyticsService` and return it from the constructors instead of returning an interface (e.g. `Service`)

internal/analytics/analytics.go
```diff
@@ -0,0 +1,119 @@
...
+		distinctID:       distinctID,
+		startupTime:      time.Now().Unix(),
+		client:           client,
+		isAura:           isAura,
+	}
+
+	return &analytics{cfg: cfg, disabled: false}
+}
+
+func NewAnalytics(mixPanelToken string, mixpanelEndpoint string, isAura bool) Service {
```
keremgocen: same here, it would be more explicit to return the analytics service implementation (a pointer to it rather), instead of the interface

.golangci.yaml
keremgocen: I can see some of the naked returns and how they can be useful now in specific scenarios, but I'm not happy with the idea of enabling it globally and not seeing any warnings for other naked returns. So please let's do inline lint ignores for those places instead of chaning the linter config. https://golangci-lint.run/docs/linters/false-positives/#nolint-directive

internal/analytics/analytics.go
MacondoExpress: Yes you have spotted the reason, we don't handle errors on EmitEvent outside printing at this line, this is by design as we don't want any of the consequences:
- handle always errors at the client code.
- fail in case of telemetry errors.
- disable error check lint rules.

.goreleaser.yaml
```diff
@@ -10,6 +10,8 @@ builds:
       - CGO_ENABLED=0
     ldflags:
       - -X "main.Version={{.Env.GORELEASER_CURRENT_TAG}}"
+      - -X "main.MixPanelEndpoint=https://api.mixpanel.com"
+      - -X "main.MixPanelToken=4bfb2414ab973c741b6f067bf06d5575"
```
keremgocen, line 14: I think we already discussed this being public, but just to clarify, is there an alternative way to prevent people from getting this token and populating our metrics with malformed data?

mjfwebb: One alternative here would be to have an additional API where we simply accept a http POST of something like this:

```
{
 "uid": "abc123",
 "event": "eventName"
)
```

On the other hand, it's quite common to have analytics tokens available on the user side. Their own documentation suggests the token will be in view of the users.

internal/analytics/interfaces.go
keremgocen: I'm taking this back based on my own earlier recommendation. If we did this, it would read as `analytics.Analytics`. I think we can keep it as `Service`, what's more important is to return the implementation instead of the interface here I think: `func NewAnalyticsWithClient(mixPanelToken string, mixpanelEndpoint string, client HTTPClient, isAura bool) Service {`

internal/analytics/analytics.go
keremgocen: I was also wondering if telemetry events should be sent in their own goroutine in parallel to the rest of the execution, but if we are sure nothing is crashing and burning in case of any telemetry errors, we should be ok for now. (i.e. telemetry errors should not crash the main app)

internal/analytics/analytics.go
keremgocen: thinking about this again, I almost want to inline it as `distinctID, err := uuid.NewV6()` where it's used, but I guess you CAN get errors from that in theory. And we want to log it, so this should be ok. It becomes a question of whether we want to log it where it's used or in a tidy function here. I'm resolving my comment:)

internal/config/config.go
```diff
@@ -7,18 +7,23 @@ import (
...
+	ReadOnly  string // If true, disables write tools
+	Telemetry string // if false, disables telemetry
 }
 
 // Validate validates the configuration and returns an error if invalid
 func (c *Config) Validate() error {
 	if c == nil {
 		return fmt.Errorf("configuration is required but was nil")
 	}
+	// TODO: support different Config types with validation https://linear.app/neo4j/issue/DEVSURF-873/support-different-types-in-the-mcp-config
```
keremgocen: can we also remove the todo please (linter won't be happy with it either)?
we already have a ticket to track it and no one will come back and read the todo (except the linter)

internal/database/service_test.go
```diff
@@ -6,7 +6,7 @@ import (
 	"testing"
 
 	"github.com/neo4j/mcp/internal/database"
-	"github.com/neo4j/mcp/internal/database/mocks"
+	database_mocks "github.com/neo4j/mcp/internal/database/mocks"
```
keremgocen: nit: should we stick to the convention for package names so it reads a bit better https://go.dev/doc/effective_go#package-names

`database_mocks.NewMockService(ctrl)` reads a bit weird imo
my suggestion is we can important them as `db` or even `database` so it reads like this in tests:
```
database.NewMockService(ctrl) or db.NewMockService(ctrl)
analytics.NewMockService(ctrl)
```
just anything but snake case please:)
note: file names are ok, for some reason

internal/server/server_test.go
```diff
@@ -3,8 +3,9 @@ package server_test
 import (
 	"testing"
 
+	analytics_mocks "github.com/neo4j/mcp/internal/analytics/mocks"
 	"github.com/neo4j/mcp/internal/config"
-	"github.com/neo4j/mcp/internal/database/mocks"
+	db_mock "github.com/neo4j/mcp/internal/database/mocks"
```
keremgocen, line 8: Just as I was writing about the snake case, this is also inconsistent with the rest `db_mock` vs `database_mocks`

test/integration/helpers/helpers.go
```diff
@@ -58,14 +60,31 @@ func NewTestContext(t *testing.T, driver *neo4j.DriverWithContext) *TestContext
...
+	deps := &tools.ToolDependencies{DBService: databaseService, AnalyticsService: analyticsService}
 
 	tc.Service = databaseService
 	tc.Deps = deps
 
 	return tc
 }
 
+// getAnalyticsMock is used to mock the analytics service, for integration test purpose.
+func getAnalyticsMock(t *testing.T) *analytics_mocks.MockService {
```
keremgocen, line 73: I think we already discussed whether disabling this could be an option instead. 

But I'm also thinking, since these are integration tests, what's the downside of having actual analytics being fired from the tests? Do you think we can have that in an isolated namespace/API key, or would it be too expensive?

internal/analytics/analytics.go
MacondoExpress: Absolutely good point!

internal/analytics/analytics_test.go
MacondoExpress: Inside the "events.go", 
eventually we could move these to be constants, but as I don't foresee I case for these to be reused, I think it improve the readability for the next guy having inline when they are needed, instead of create a layer of indirection IMO, no strongly opinionated to be fair, but I slightly prefer this way

internal/analytics/events.go
MacondoExpress: Commented above, on the test line, no problem applying this suggestion if you strongly disagree
> Inside the "events.go",
> eventually we could move these to be constants, but as I don't foresee I case for these to be reused, I think it improve the readability for the next guy having inline when they are needed, instead of create a layer of indirection IMO, no strongly opinionated to be fair, but I slightly prefer this way

test/integration/helpers/helpers.go
MacondoExpress: Yes, we could discuss setting up a test service for MixPanel, but we ended up out of scope for now. 
I think it's mainly the timeline not matching more than the cost btw

internal/analytics/events.go
keremgocen: I think we are good for now since the `eventNamePrefix` makes it a lot more obvious. Initially, I was leaning towards something like this:
```
// internal/analytics/events.go
const (
    EventGDSProjCreated = "GDS_PROJ_CREATED"
    EventGDSProjDrop    = "GDS_PROJ_DROP"
    EventMCPStartup     = "MCP_STARTUP"
    EventToolUsed       = "TOOL_USED"
)

func (a *Analytics) NewGDSProjCreatedEvent() TrackEvent {
      return TrackEvent{
          Event:      eventNamePrefix + "_" + EventGDSProjCreated,
          Properties: getBaseProperties(a.cfg),
      }
  }
```

Alternatively, we could gather all events in one file and use it from there, so we don't have factory methods. Another option could be grouping them by domain, but it will depend on how much it will grow over time. I reckon we might want to have more eventually, but it's currently only a cosmetic change, would not really add much if we change anything now imo. So I'm happy with the current state.

PR #93, opened 2025-11-13
Implement unified logger with log level and redaction

keremgocen:
Implements a unified logging system for the Neo4j MCP server using Go's standard `log/slog` package. Replaces scattered `log.Printf` calls throughout the codebase with structured logging and adds support for MCP's `setLevel` notification to adjust log levels at runtime dynamically.

Part 1: (in this PR)
- covers the foundations, uses a global slog instance with log levels from the MCP spec
- An MCP server hook allows MCP clients to set the level
- log format and output are decided at the start via env variables (although VSCode MCP client sets it to INFO, and I could not find a way to change it)
- added basic test coverage

Part 2: (not in this PR but here #95 -- to be updated after the global slog changes)
- Add a redaction mechanism to remove sensitive data
- Add test coverage

  
Two new environment variables control logging behaviour:
- `NEO4J_LOG_LEVEL` (default: "info") - Sets the initial log level
- `NEO4J_LOG_FORMAT` (default: "text") - Sets output format ("text" or "json")
Both values are validated at startup with helpful warnings if invalid.

### DIFF
--- a/cmd/neo4j-mcp/main.go
+++ b/cmd/neo4j-mcp/main.go
@@ -57,11 +57,7 @@ func main() {
 		return
 	}
 
-	anService, err := analytics.NewAnalytics(MixPanelToken, MixPanelEndpoint, cfg.URI)
-	if err != nil {
-		slog.Error("Failed to create analytics service", "error", err)
-		return
-	}
+	anService := analytics.NewAnalytics(MixPanelToken, MixPanelEndpoint, cfg.URI)
 
 	if cfg.Telemetry == "false" || MixPanelEndpoint == "" || MixPanelToken == "" {
 		slog.Info("Telemetry disabled.")
--- a/internal/analytics/analytics.go
+++ b/internal/analytics/analytics.go
@@ -31,7 +31,7 @@ type Analytics struct {
 }
 
 // for testing purposes - enables dependency injection of http client
-func NewAnalyticsWithClient(mixPanelToken string, mixpanelEndpoint string, client HTTPClient, uri string) (*Analytics, error) {
+func NewAnalyticsWithClient(mixPanelToken string, mixpanelEndpoint string, client HTTPClient, uri string) *Analytics {
 	distinctID := getDistinctID()
 	cfg := analyticsConfig{
 		token:            mixPanelToken,
@@ -42,10 +42,10 @@ func NewAnalyticsWithClient(mixPanelToken string, mixpanelEndpoint string, clien
 		isAura:           isAura(uri),
 	}
 
-	return &Analytics{cfg: cfg, disabled: false}, nil
+	return &Analytics{cfg: cfg, disabled: false}
 }
 
-func NewAnalytics(mixPanelToken string, mixpanelEndpoint string, uri string) (*Analytics, error) {
+func NewAnalytics(mixPanelToken string, mixpanelEndpoint string, uri string) *Analytics {
 	distinctID := getDistinctID()
 	cfg := analyticsConfig{
 		token:            mixPanelToken,
@@ -56,7 +56,7 @@ func NewAnalytics(mixPanelToken string, mixpanelEndpoint string, uri string) (*A
 		isAura:           isAura(uri),
 	}
 
-	return &Analytics{cfg: cfg, disabled: false}, nil
+	return &Analytics{cfg: cfg, disabled: false}
 }
 
 func isAura(uri string) bool {
--- a/internal/analytics/analytics_test.go
+++ b/internal/analytics/analytics_test.go
@@ -13,14 +13,10 @@ import (
 	"go.uber.org/mock/gomock"
 )
 
-// newTestAnalytics creates an analytics service for testing and fails the test if creation fails
+// newTestAnalytics creates an analytics service for testing
 func newTestAnalytics(t *testing.T, token, endpoint string, client analytics.HTTPClient, uri string) *analytics.Analytics {
 	t.Helper()
-	service, err := analytics.NewAnalyticsWithClient(token, endpoint, client, uri)
-	if err != nil {
-		t.Fatalf("failed to create analytics service: %v", err)
-	}
-	return service
+	return analytics.NewAnalyticsWithClient(token, endpoint, client, uri)
 }
 
 func TestAnalytics(t *testing.T) {


cmd/neo4j-mcp/main.go
```diff
@@ -20,49 +22,55 @@ func main() {
...
 	cfg, err := config.LoadConfig()
 	if err != nil {
-		log.Fatalf("Failed to load configuration: %v", err)
+		// Can't use logger here yet, so just print to stderr
+		fmt.Println("Failed to load configuration: " + err.Error())
+		os.Exit(1)
 	}
 
+	// Initialize logger
+	log := logger.New(cfg.LogLevel, cfg.LogFormat, os.Stderr)
```
keremgocen: Initially we can set the level, format and the output ~~stream~~,  as long as it implements the `io.Writer` interface.

internal/logger/logger_test.go
```diff
@@ -0,0 +1,163 @@
...
+		if !strings.Contains(output, "error message mixed case") {
+			t.Error("Expected Error (mixed case) to change log level to error")
+		}
+		if strings.Contains(output, "info should not appear") {
+			t.Error("Expected info to NOT appear at error level")
+		}
+	})
+
+	t.Run("all log levels can be set dynamically", func(t *testing.T) {
+		levels := []string{"debug", "info", "notice", "warn", "warning", "error", "critical", "alert", "emergency"}
```
keremgocen: we probably won't need this many levels, but it's in the spec https://modelcontextprotocol.io/specification/draft/server/utilities/logging#log-levels

internal/tools/cypher/read_cypher_handler.go
```diff
@@ -2,85 +2,85 @@ package cypher
...
-		log.Printf("Error binding arguments: %v", err)
+		deps.Log.Error("error binding arguments", "error", err)
 		return mcp.NewToolResultError(err.Error()), nil
 	}
 
 	Query := args.Query
 	Params := args.Params
 
-	log.Printf("cypher-query: %s", Query)
+	deps.Log.Info("executing read cypher query", "query", Query)
```
keremgocen: I'd keep this as a `debug` instead, but VSCode MCP client is setting the log level to `Info` and I could not find a way to change that. Not sure if we want cypher queries to be logged all the time?

.gitignore
```diff
@@ -36,3 +36,7 @@ go.work.sum
 bin/
 # Added by goreleaser init:
 dist/
+
+# Agents
+CLAUDE.md
+gemini.md
```
MacondoExpress: Missing new line, also you may want to add these to your .git/info/exclude

internal/logger/logger.go
```diff
@@ -0,0 +1,106 @@
...
+	service := &Service{
+		Logger: slog.New(handler),
+		level:  levelVar,
+	}
+
+	return service
+}
+
+const (
+	LevelNotice    = slog.Level(2)  // Between Info and Warn
```
MacondoExpress: We could not export these

internal/config/config.go
```diff
@@ -52,6 +54,8 @@ func LoadConfig() (*Config, error) {
 		Database:  GetEnvWithDefault("NEO4J_DATABASE", "neo4j"),
 		ReadOnly:  GetEnvWithDefault("NEO4J_READ_ONLY", "false"),
 		Telemetry: GetEnvWithDefault("NEO4J_TELEMETRY", "true"),
+		LogLevel:  GetEnvWithDefault("LOG_LEVEL", "info"),
```
MacondoExpress: I would change it to have it "NEO4J_" as prefix to avoid collision with other user's environment variable

internal/config/config.go
```diff
@@ -52,6 +54,8 @@ func LoadConfig() (*Config, error) {
...
+		LogFormat: GetEnvWithDefault("LOG_FORMAT", "text"),
```
MacondoExpress: Let's consider if we want to expose this, in the case we expose it, we should validate it !

internal/logger/logger.go
```diff
@@ -0,0 +1,106 @@
...
+	// Create the logger service
+	service := &Service{
+		Logger: slog.New(handler),
+		level:  levelVar,
+	}
+
+	return service
+}
+
+const (
```
MacondoExpress: Nitpick: You can add a comment with the link: https://pkg.go.dev/log/slog#Level to reduce the time to get the info for the reader.

internal/logger/logger.go
```diff
@@ -0,0 +1,106 @@
...
+		Level:       levelVar,
+		ReplaceAttr: replaceAttr,
+	}
+
+	var handler slog.Handler
+	switch strings.ToLower(format) {
+	case "json":
+		handler = slog.NewJSONHandler(writer, opts)
+	default:
+		handler = slog.NewTextHandler(writer, opts)
```
MacondoExpress, line 88: I would handle the log format/level explicitly and logging the what the user passed as environment variable, right now, if you pass an invalid "LOG_FORMAT" or "LOG_LEVEL" the user has no visibility of the mistake!

internal/logger/logger.go
```diff
@@ -0,0 +1,106 @@
...
+	case "alert":
+		return LevelAlert
+	case "emergency":
+		return LevelEmergency
+	default:
+		return slog.LevelInfo
+	}
+}
+
+func replaceAttr(_ []string, a slog.Attr) slog.Attr {
```
MacondoExpress, line 152: From the reader perspective is not clear the intention of why we need the `replaceAttr` func rather than relying on the slog formatter.
From what I guess, it's because slog convert levels to string using a small subset of values:
- DEBUG
- INFO
- WARN
- ERROR

Giving back less granularity than what we desire, but as it's not tested the underline reason could be lost over time!

internal/server/server.go
```diff
@@ -17,31 +17,42 @@ type Neo4jMCPServer struct {
...
+
 	mcpServer := server.NewMCPServer(
 		"neo4j-mcp",
 		version,
 		server.WithToolCapabilities(true),
 		server.WithInstructions("This is the Neo4j official MCP server and can provide tool calling to interact with your Neo4j database,"+
 			"by inferring the schema with tools like get-schema and executing arbitrary Cypher queries with read-cypher."),
+		server.WithLogging(),
+		server.WithHooks(&server.Hooks{
+			OnAfterSetLevel: []server.OnAfterSetLevelFunc{srv.onAfterSetLevelHook},
```
MacondoExpress: ❤️

MacondoExpress (commented):
Really good! Left some comments, nothing of these are massive blocker IMO

internal/config/config.go
keremgocen: not sure I understand this comment, do you want to validate the log format? why?

internal/logger/logger.go
keremgocen: I'm not sure typos in the log level or format is as critical as defaulting the username password for example. Defaulting to info and text format sounded like a reasonable enough option to me.
Do you think we should error out if that happens instead of using defaults?

internal/config/config.go
MacondoExpress: I'll reply on the comment below

internal/logger/logger.go
MacondoExpress: I think we've ended up choosing the warns and default option!

internal/logger/logger.go
keremgocen: I added a doc string to explain what `replaceAttr` is used for. We also have a test already `all log levels can be set dynamically` and now we are checking for valid values so hopefully it should be a lot more clear.

internal/tools/cypher/write_cypher_handler.go
```diff
@@ -2,72 +2,71 @@ package cypher
...
+func handleWriteCypher(ctx context.Context, request mcp.CallToolRequest, deps *tools.ToolDependencies) (*mcp.CallToolResult, error) {
+	// Emit analytics event
+	if deps.AnalyticsService != nil {
+		deps.AnalyticsService.EmitEvent(deps.AnalyticsService.NewToolsEvent("write-cypher"))
 	}
-	if dbService == nil {
-		errMessage := "Database service is not initialized"
-		log.Printf("%s", errMessage)
-		return mcp.NewToolResultError(errMessage), nil
-	}
```
mjfwebb, line 30: Why was this moved down, I believe we had it further up to fail fast.

keremgocen: It must have been moved when I was fixing the merge conflicts, it was not intentional. Good catch, I'll move it back.

keremgocen: Looks like I nuked the analytics service nil check as well, it's back now

PR #96, opened 2025-11-17
feat: Implement startup requirement checks and dynamic tool registration

MacondoExpress:
This update introduces a comprehensive pre-flight checks during the startup sequence. 
We've implemented a new `verifyRequirements` function to ensure that the server's environment meets all necessary prerequisites before it begins its main operations.

The verification process now confirms the following:
- Valid connection to the configured Neo4j instance.
- Ability to perform a query against the configured instance.
- Verifies correctness of the APOC installation.
- In case of a missing GDS requirement, the server will start without the GDS specific tools.

To perform these, this PR includes also other improvements such as the ability to filter registered tools by their specific categories, for instance: "read-only", "cypher", "gds".

The `registerTools` method is no longer exposed but instead is been moved to be part of the `.Start` method and implicitly invoked.

Integration test helpers are slightly changed to accomodates testing initial part of the Server initialisation.


internal/server/server_test.go
```diff
@@ -1,12 +1,14 @@
 package server_test
 
 import (
+	"fmt"
 	"testing"
 
 	analytics "github.com/neo4j/mcp/internal/analytics/mocks"
 	"github.com/neo4j/mcp/internal/config"
-	db_mock "github.com/neo4j/mcp/internal/database/mocks"
+	db "github.com/neo4j/mcp/internal/database/mocks"
```
keremgocen, line 9: 🫶

internal/server/tool_register_test.go
```diff
@@ -1,41 +1,43 @@
 package server_test
 
 import (
+	"fmt"
 	"testing"
 
-	"github.com/neo4j/mcp/internal/analytics"
-	analytics_mock "github.com/neo4j/mcp/internal/analytics/mocks"
+	amock "github.com/neo4j/mcp/internal/analytics/mocks"
```
keremgocen: I think we are not using the real `analytics` here, so there should not be a conflict if we used `analytics` instead of `amock` to be a bit more specific.

test/integration/helpers/helpers.go
```diff
@@ -30,12 +30,13 @@ func (ul UniqueLabel) String() string {
...
-	Service       database.Service
-	Deps          *tools.ToolDependencies
-	createdLabels map[string]bool
+	ctx              context.Context
+	t                *testing.T
+	TestID           string
+	Service          database.Service
+	Deps             *tools.ToolDependencies
+	createdLabels    map[string]bool
+	AnalyticsService *analytics_mocks.MockService
```
keremgocen: Same here, I guess it was done in an earlier PR, but I think we should use `analytics` here as well. At least name it something other than snake case, please:)

internal/database/service.go
```diff
@@ -28,9 +28,18 @@ func NewNeo4jService(driver neo4j.DriverWithContext, database string) (*Neo4jSer
 	}, nil
 }
 
+// VerifyConnectivity checks the driver can establish a valid connection with a Neo4j instance;
+func (s *Neo4jService) VerifyConnectivity(ctx context.Context) error {
```
keremgocen, line 32: this looks neat as part of the service now

internal/server/server.go
```diff
@@ -31,30 +33,91 @@ func NewNeo4jMCPServer(version string, cfg *config.Config, dbService database.Se
...
-		config:    cfg,
-		dbService: dbService,
-		version:   version,
-		anService: anService,
+		MCPServer:    mcpServer,
+		config:       cfg,
+		dbService:    dbService,
+		version:      version,
+		anService:    anService,
+		gdsInstalled: false,
```
keremgocen, line 41: nit: no need to provide it, as it defaults to `false`
Counterargument: it's more visible to be explicit

MacondoExpress: I think it will be more visible for no harm introduced, I will be prone to keep it, agree?

keremgocen (approved):
LGTM

PR #97, opened 2025-11-18
Add Dockerfile

mjfwebb:
In preparation for adding neo4j-mcp to the Docker mcp-register, we need to have a Dockerfile present in our repository.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,5 +26,8 @@ WORKDIR /app
 # Copy binary from builder
 COPY --from=builder /build/neo4j-mcp /app/neo4j-mcp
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+    -o ../../neo4j-mcp
+
+# Runtime stage
+FROM scratch
+
+WORKDIR /app
+
+# Copy binary from builder
+COPY --from=builder /build/neo4j-mcp /app/neo4j-mcp
+
```
luffy1727, line 28: maybe specify the user here as a non root user USER 65532:65532

PR #100, opened 2025-11-18
Remove defaults and add manual flag parsing for configuration

keremgocen:
Introduce manual parsing for CLI flags related to Neo4j configuration, allowing users to override environment variables. Enhance test coverage for edge cases and integration tests with real environment variables.

### DIFF
--- a/.changes/unreleased/Minor-20251120-144649.yaml
+++ b/.changes/unreleased/Minor-20251120-144649.yaml
@@ -1,0 +1,3 @@
+kind: Minor
+body: Added CLI flags for configuration (--neo4j-uri, --neo4j-username, --neo4j-password, --neo4j-database, --neo4j-read-only, --neo4j-telemetry) that override environment variables. NEO4J_URI, NEO4J_USERNAME, and NEO4J_PASSWORD are now required environment variables (no defaults provided).
+time: 2025-11-20T14:46:49.67341Z


internal/cli/args.go
```diff
@@ -14,38 +15,101 @@ Usage:
...
 
 Options:
-  -h, --help      Show this help message
-  -v, --version   Show version information
+  -h, --help                          Show this help message
+  -v, --version                       Show version information
+  --neo4j-uri <URI>                   Neo4j connection URI (overrides env var)
+  --neo4j-username <USERNAME>         Database username (overrides env var)
+  --neo4j-password <PASSWORD>         Database password (overrides env var)
+  --neo4j-database <DATABASE>         Database name (overrides env var)
```
mjfwebb: I think I would like to be a bit more explicit here

```suggestion
  --neo4j-uri <URI>                   Neo4j connection URI (overrides environment variable NEO4J_URI)
  --neo4j-username <USERNAME>         Database username (overrides environment variable NEO4J_USERNAME)
  --neo4j-password <PASSWORD>         Database password (overrides environment variable NEO4J_PASSWORD
  --neo4j-database <DATABASE>         Database name (overrides environment variable NEO4J_DATABASE)

```

internal/cli/args.go
```diff
@@ -14,38 +15,101 @@ Usage:
...
-  NEO4J_PASSWORD  Database password (required)
+Required Environment Variables:
+  NEO4J_URI       Neo4j database URI
+  NEO4J_USERNAME  Database username
+  NEO4J_PASSWORD  Database password
+
+Optional Environment Variables:
   NEO4J_DATABASE  Database name (default: neo4j)
+  NEO4J_TELEMETRY Enable/disable telemetry (default: true)
+  NEO4J_READ_ONLY Enable read-only mode (default: false)
```
mjfwebb, line 36: We don't want these to be args?

internal/cli/args.go
```diff
@@ -14,38 +15,101 @@ Usage:
...
+- os.Args[4] = "test"
+
+As the loop processes:
+1. i=1: Matches case "--neo4j-uri" → i += 2 → i=3 (skips the URI value)
+2. i=3: Matches case "--neo4j-username" → i += 2 → i=5 (skips the "test" value)
+3. i=5: Loop ends
+
+This allows the arguments to "pass through" untouched so that flag.Parse() in main.go can later handle them properly.
+*/
+
```
mjfwebb: Maybe it's just me but I find this unnecessarily verbose, when the code itself also explains how this works.

cmd/neo4j-mcp/main.go
```diff
@@ -19,12 +20,37 @@ var MixPanelEndpoint = ""
 var MixPanelToken = ""
 
 func main() {
+	// Define CLI flags for configuration
```
mjfwebb: Just a personal choice here, but I would have likely put all of these code additions into `config.LoadConfig()` instead of polluting `main()`

mjfwebb (commented):
Very nice! Left some comments for your consideration 👍

cmd/neo4j-mcp/main.go
MacondoExpress: +1

cmd/neo4j-mcp/main.go
```diff
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"flag"
```
MacondoExpress: what a nice std library

cmd/neo4j-mcp/main.go
```diff
@@ -19,12 +20,37 @@ var MixPanelEndpoint = ""
...
-	cfg, err := config.LoadConfig()
-	if err != nil {
+	// Parse flags after HandleArgs
+	flag.Parse()
+
+	// Load config from environment variables
+	cfg := config.LoadConfig()
+
+	// Override config with CLI flags if provided (CLI flags take precedence)
+	if *neo4jURI != "" {
```
MacondoExpress: Same as @mjfwebb pointed out, I would delegate this "overwrite" to the config package, same as "cfg.Validate()"  I see no benefits to delegate the consumer to correctly invoke it, I would delegate that method to be consumed by `config.loadConfig`

cmd/neo4j-mcp/main.go
MacondoExpress: IMO, here we should delegate the "cli" and "config" with clear separation of ownership, from the main.go file I would keep the consumer logic as dummy as possible:
- "handleArgs" we could return a typed struct with the expected args so that can be used later by "loadConfig"
- "loadConfig" that returns errors in case of validation errors

test/integration/config/config_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+// TestLoadConfig_Neo4jConnectivity tests that the application can connect to Neo4j with the provided configuration.
+// This test requires the following environment variables to be set:
+//   - NEO4J_URI: Neo4j database URI
+//   - NEO4J_USERNAME: Database username
+//   - NEO4J_PASSWORD: Database password
+//
+// If any of these environment variables are missing, the test will fail with a clear error message.
+// If the environment variables are set but the database is not accessible, the test will fail with a connection error.
+func TestLoadConfig_Neo4jConnectivity(t *testing.T) {
```
MacondoExpress: Let's remove this test, as we can test entirely with unit testing, without undesired side effects!

internal/cli/args.go
keremgocen: converting them to args as well, I think it makes sense

internal/cli/args.go
keremgocen: done

cmd/neo4j-mcp/main.go
keremgocen: moved the flag parsing logic into LoadConfig

cmd/neo4j-mcp/main.go
keremgocen: This should be done now, handleArgs is returning a struct to override the env vars from the cli flag parsing
loadConfig returns errors if validation fails

cmd/neo4j-mcp/main.go
```diff
@@ -50,10 +60,10 @@ func main() {
 
 	anService := analytics.NewAnalytics(MixPanelToken, MixPanelEndpoint, cfg.URI)
 
-	if cfg.Telemetry == "false" || MixPanelEndpoint == "" || MixPanelToken == "" {
+	if !cfg.Telemetry || MixPanelEndpoint == "" || MixPanelToken == "" {
 		log.Println("Telemetry disabled.")
 		anService.Disable()
-	} else if cfg.Telemetry == "true" {
+	} else {
 		anService.Enable()
```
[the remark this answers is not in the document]
keremgocen: reverted the logic so that we only enable telemetry when all conditions are met at the same time

internal/cli/args.go
```diff
@@ -14,38 +16,116 @@ Usage:
...
+				err = fmt.Errorf("%s requires a value (got flag %s instead)", arg, nextArg)
+				break
+			}
+			// Safe to skip flag and value - let flag package handle them
+			i += 2
 		default:
-			err = fmt.Errorf("unknown flag or argument: %s", arg)
+			if arg == "--" {
+				// Stop processing our flags, let flag package handle the rest
+				i = len(os.Args)
```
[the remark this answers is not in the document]
keremgocen: It doesn't work correctly in this context (it only breaks out of the switch, not the loop). 
We have two valid options:
1. Keep `i = len(os.Args)` (current) - Works correctly, though less conventional
2. Use labelled break - More idiomatic Go, but adds complexity

I think I'd like to stick to the `i = len(os.Args)` instead of a labelled break, which ends the loop and the code works correctly.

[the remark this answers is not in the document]
mjfwebb: No strong opinions here. I think either way is fine, though I could make the argument that a labelled break is quicker to understand.

internal/cli/args.go
```diff
@@ -14,38 +16,116 @@ Usage:
...
+		// Allow configuration flags to be parsed by the flag package
+		case "--neo4j-uri", "--neo4j-username", "--neo4j-password", "--neo4j-database", "--neo4j-read-only", "--neo4j-telemetry":
+			// Check if there's a value following the flag
+			if i+1 >= len(os.Args) {
+				err = fmt.Errorf("%s requires a value", arg)
+				break
+			}
+			// Check if next argument is another flag (starts with --)
+			nextArg := os.Args[i+1]
+			if strings.HasPrefix(nextArg, "-") {
```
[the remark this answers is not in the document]
keremgocen: 😭 you just suggested the `--` earlier:)

internal/cli/args.go
```diff
@@ -14,38 +16,116 @@ Usage:
...
+			i++
 		case "-v", "--version":
 			flags["version"] = true
+			i++
+		// Allow configuration flags to be parsed by the flag package
+		case "--neo4j-uri", "--neo4j-username", "--neo4j-password", "--neo4j-database", "--neo4j-read-only", "--neo4j-telemetry":
+			// Check if there's a value following the flag
+			if i+1 >= len(os.Args) {
+				err = fmt.Errorf("%s requires a value", arg)
+				break
```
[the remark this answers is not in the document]
keremgocen: we exit the loop on #L127 if there's an error in the switch so that's incorrect

internal/cli/args.go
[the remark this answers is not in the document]
keremgocen: We decided to revisit this at a later stage and think about whether we want to use some helper packages similar to in aura-cli for the flags.

PR #104, opened 2025-11-20
Fix: Correctly parse floats in Cypher parameters

MacondoExpress:
This change addresses an issue where numbers with decimal parts (e.g., 1.0) provided to the read-cypher and write-cypher commands were incorrectly processed as integers.

By implementing a custom JSON unmarshaler for CypherParams to differentiate between integers and floating-point numbers during deserialization. The new logic attempts to parse a number as an int64, and if that fails due to a decimal point, it parses it as a float64.

Compared to the previous iteration, this iteration:
- Is implemented directly on the Go type, by implementing the "Unmarshaler" interface..
- Converts 1.0 to float64 instead of int64

### DIFF
--- a/internal/tools/cypher/json_tagged_types.go
+++ b/internal/tools/cypher/json_tagged_types.go
@@ -5,9 +5,9 @@ import (
 	"encoding/json"
 )
 
-type CypherParams map[string]any
+type Params map[string]any
 
-func (cp *CypherParams) UnmarshalJSON(data []byte) error {
+func (cp *Params) UnmarshalJSON(data []byte) error {
 	decoder := json.NewDecoder(bytes.NewReader(data))
 
 	decoder.UseNumber()
--- a/internal/tools/cypher/json_tagged_types_test.go
+++ b/internal/tools/cypher/json_tagged_types_test.go
@@ -15,50 +15,50 @@ func TestBindArgumentsWithReadCypherInput(t *testing.T) {
 		name       string
 		arguments  map[string]any
 		wantQuery  string
-		wantParams cypher.CypherParams
+		wantParams cypher.Params
 		wantErr    bool
 	}{
 		{
 			name: "basic integer parameter",
 			arguments: map[string]any{
 				"query":  "MATCH (n) WHERE n.id = $id RETURN n",
-				"params": cypher.CypherParams{"id": 1},
+				"params": cypher.Params{"id": 1},
 			},
 			wantQuery:  "MATCH (n) WHERE n.id = $id RETURN n",
-			wantParams: cypher.CypherParams{"id": int64(1)},
+			wantParams: cypher.Params{"id": int64(1)},
 		},
 		{
 			name: "basic float parameter",
 			arguments: map[string]any{
 				"query":  "MATCH (n) WHERE n.value = $value RETURN n",
-				"params": cypher.CypherParams{"value": 1.5},
+				"params": cypher.Params{"value": 1.5},
 			},
 			wantQuery:  "MATCH (n) WHERE n.value = $value RETURN n",
-			wantParams: cypher.CypherParams{"value": float64(1.5)},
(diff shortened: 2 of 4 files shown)


internal/tools/cypher/json_tagged_types.go
```diff
@@ -0,0 +1,50 @@
...
+	decoder := json.NewDecoder(bytes.NewReader(data))
+
+	decoder.UseNumber()
+
+	var temp map[string]any
+	if err := decoder.Decode(&temp); err != nil {
+		return err
+	}
+
+	*cp = ConvertNumbers(temp).(map[string]any)
```
keremgocen, line 20: This does not feel right to me at the first glance, why are we doing it?

keremgocen: since we are doing type assertion here, it would be nice to handle the error as it can panic runtime if `ConvertNumbers()` somehow returns a non-map type. While unlikely given the input is always a map, we can return something useful like "unexpected type after conversion". 
The other issue is `ConvertNumbers` both mutates AND returns the same value, which is why it reads a bit weird I think.
Go functions typically either:
    - Mutate in place and return nothing (like `sort.Strings()`)
    - Return a new value without mutating (like `strings.ToUpper()`)
    - not both

What I'd like to propose is to expand this logic a bit (we can do it together if you prefer) so we either mutate the value or return new maps/slices, otherwise it's harder to maintain imo.

internal/tools/cypher/json_tagged_types.go
```diff
@@ -0,0 +1,50 @@
...
+
+func ConvertNumbers(input any) any {
+	switch v := input.(type) {
+	case json.Number:
+		// Try to parse as Int64 first
+		if i, err := v.Int64(); err == nil {
+			return i
+		}
+		// If it fails (because of decimal point), parse as Float64
+		if f, err := v.Float64(); err == nil {
```
keremgocen, line 32: I think this works nicely because golang parses "1.0" by default as int64 but if do `decoder.UseNumber()` that initial attempt to convert to int fails and we can parse it as float right?

internal/tools/cypher/json_tagged_types.go
```diff
@@ -0,0 +1,50 @@
+package cypher
+
+import (
+	"bytes"
+	"encoding/json"
+)
+
+type Params map[string]any
```
keremgocen, line 8: It would be nice to add godoc comments for exported types and functions so this is easier to understand, something like: (maybe also for the Unmarshall function)


```suggestion
// Params is a map of Cypher query parameters with custom JSON unmarshaling
// that preserves numeric types correctly for Neo4j.
//
// When unmarshaling from JSON:
//   - Whole numbers (e.g., 1, 42, -10) become int64
//   - Numbers with fractional parts (e.g., 1.5, 3.14) become float64
//   - Numbers with decimal notation but no fraction (e.g., 10.0) become float64
//   - Other types (strings, booleans, null) are preserved as-is
type Params map[string]any
```

internal/tools/cypher/json_tagged_types.go
MacondoExpress: Yes, when we use "useNumber" it will be converted to an interface "number" and "1.0" will fail, and it will be converted to float

internal/tools/cypher/json_tagged_types.go
MacondoExpress: Done the fix for the panic error, 
I'll open the PR, 
The "ConvertNumbers" I left as it was from your PR, if you look "params_converter.go" 
is the same method you have added with the difference for the float handling, but that's not changes the "mutate and return", let's not extend the scope of this improvement. 
https://github.com/neo4j/mcp/pull/104/files/227ce17e3c5d870a5b95f8b1e84b015442c311be#diff-7be318cc07d6bbd792cb31b3d946e940751f2180f75673f4207f73591589e416L54

PR #115, opened 2025-11-27
Add transport mode configuration and basic HTTP server

keremgocen:
Introduce a transport mode configuration for the server, enabling both standard input/output and HTTP modes. Implement a basic HTTP server alongside updates to the configuration and tests to ensure proper functionality.

Note: this version is to be merged into a separate branch while we continue working on the next bit for HTTP MCP, that is, middlewares for auth and allowed origins, plus the TLS support. I'd like to keep the PRs smaller to make them easy to review.

### DIFF
--- a/test/integration/dbservice/dbservice.go
+++ b/test/integration/dbservice/dbservice.go
@@ -62,21 +62,16 @@ func (dbs *dbService) GetDriver() *neo4j.DriverWithContext {
 	return dbs.driver
 }
 
-func (dbs *dbService) GetDriverConf(transport string) *config.Config {
+func (dbs *dbService) GetDriverConf() *config.Config {
 	if dbs.useContainer == true {
 		return containerrunner.GetDriverConf()
-	}
-
-	// Default to stdio transport if not specified
-	if transport == "" {
-		transport = config.TransportModeStdio
 	}
 
 	cfg := &config.Config{
 		URI:           config.GetEnvWithDefault("NEO4J_URI", "bolt://localhost:7687"),
 		Username:      config.GetEnvWithDefault("NEO4J_USERNAME", "neo4j"),
 		Password:      config.GetEnvWithDefault("NEO4J_PASSWORD", "password"),
-		TransportMode: config.GetEnvWithDefault("NEO4J_MCP_TRANSPORT", transport),
+		TransportMode: config.GetEnvWithDefault("NEO4J_MCP_TRANSPORT", config.TransportModeStdio),
 	}
 
 	return cfg
--- a/test/integration/server_test.go
+++ b/test/integration/server_test.go
@@ -17,7 +17,7 @@ import (
 
 func TestServerLifecycle(t *testing.T) {
 	t.Parallel()
-	testCFG := dbs.GetDriverConf(config.TransportModeStdio)
+	testCFG := dbs.GetDriverConf()
 	testCases := []struct {
 		name        string
 		config      *config.Config


internal/server/server_http_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+	// Start server in background (it will block on ListenAndServe)
+	errChan := make(chan error, 1)
+	go func() {
+		errChan <- srv.Start()
+	}()
+
+	// Wait briefly for server to initialize (httpServer is set before ListenAndServe blocks)
+	// Relying on timing here is not ideal, we should consider adding a ready channel in the server in the future or moving this to an integration test.
+	time.Sleep(100 * time.Millisecond)
```
keremgocen, line 153: This test might be a bit brittle with all the hardcoded timeout durations, and I don't like the fact that it's time-based, as we need a few milliseconds for the server to come up. Happy to move it to an integration test later if we all feel that way.

internal/cli/args.go
```diff
@@ -97,28 +116,32 @@ func HandleArgs(version string) {
...
-		case "-h", "--help":
-			flags["help"] = true
-			i++
-		case "-v", "--version":
-			flags["version"] = true
-			i++
+
 		// Allow configuration flags to be parsed by the flag package
-		case "--neo4j-uri", "--neo4j-username", "--neo4j-password", "--neo4j-database", "--neo4j-read-only", "--neo4j-telemetry", "--neo4j-schema-sample-size":
+		if slices.Contains(argsSlice, arg) {
```
MacondoExpress, line 133: <3

internal/cli/args.go
```diff
@@ -36,6 +38,7 @@ Optional Environment Variables:
   NEO4J_TELEMETRY Enable/disable telemetry (default: true)
   NEO4J_READ_ONLY Enable read-only mode (default: false)
   NEO4J_SCHEMA_SAMPLE_SIZE Number of nodes to sample for schema inference (default: 100)
+  NEO4J_MCP_TRANSPORT MCP Transport mode (e.g., 'stdio', 'http') (default: stdio)
```
MacondoExpress, line 43: should we want to keep it as environment variable? I think we can avoid it.

cmd/neo4j-mcp/main.go
```diff
@@ -83,17 +87,33 @@ func main() {
 	// Create and configure the MCP server
 	mcpServer := server.NewNeo4jMCPServer(Version, cfg, dbService, anService)
 
-	// Gracefully handle shutdown
-	defer func() {
-		if err := mcpServer.Stop(); err != nil {
-			slog.Error("Error stopping server", "error", err)
+	if cfg.TransportMode == config.TransportModeHTTP {
```
MacondoExpress: I would handle this directly  in the Server.Stop method, I think the same benefits can be also be applied with the STDIO, and both can be harmonised.
But I'm interested also in your opinion, what do you think?
The place where I'm coming from is the desire to keep the `main func` simple and easier to read while giving the server full ownership of signal handling.

internal/server/server.go
```diff
@@ -58,9 +64,18 @@ func (s *Neo4jMCPServer) Start() error {
...
+	switch s.config.TransportMode {
+	case config.TransportModeHTTP:
+		// Note: ServeHTTP handles its own signal management for graceful shutdown
+		return s.StartHTTPServer()
+	case config.TransportModeStdio:
+		slog.Info("Starting stdio server")
+		// Note: ServeStdio handles its own signal management for graceful shutdown
+		return server.ServeStdio(s.MCPServer)
+	default:
+		return fmt.Errorf("unsupported transport mode: %s", s.config.TransportMode)
```
MacondoExpress, line 76: I was thinking to coerce in the config the transport mode and raise a validation error, but I'm not too opinionated to keep it as well!

internal/server/server.go
```diff
@@ -199,9 +214,71 @@ func recordsToStartupEventInfo(records []*neo4j.Record, mcpVersion string) analy
...
+			return err
+		}
+	}
+	slog.Info("Neo4j MCP Server stopped.")
 	return nil
 }
+
+func (s *Neo4jMCPServer) StartHTTPServer() error {
+	addr := fmt.Sprintf("%s:%s", s.config.HTTPHost, s.config.HTTPPort)
+	serverURL := fmt.Sprintf("http://%s", addr)
```
MacondoExpress: do we want to have the path `/mcp` included or configurable?

internal/server/server.go
```diff
@@ -199,9 +214,71 @@ func recordsToStartupEventInfo(records []*neo4j.Record, mcpVersion string) analy
...
+
+	// Create the StreamableHTTPServer with configuration
+	mcpServerHTTP := server.NewStreamableHTTPServer(
+		s.MCPServer,
+		server.WithStateLess(true),
+	)
+
+	// Create HTTP server with custom routes
+	mux := http.NewServeMux()
```
MacondoExpress: Is there a blocker to use: `mcpServeHTTP.start` and/or `mcpServeHTTP.WithStreamableHTTPServer`?

MacondoExpress (commented):
Left some comments!

internal/cli/args.go
keremgocen: Do you we should not have an arg version of it? Because it should be already an env variable. If so, what’s the downside of having it also as a flag? (Just trying to make sure I understand the comment)

cmd/neo4j-mcp/main.go
keremgocen: I think it’s a good point and I generally agree to keeping main simple, although it’s a common pattern for servers to have a sigterm listener in main that blocks until the exit signal is received.

I also remember there was not much to stop in the stdio version, I mean it was a blank function if I remember correctly but I will check this again in the morning:)

internal/server/server.go
keremgocen: I guess this makes it a required config right? We should probably validate we have either of the 2 possible transports and return an error was my initial reaction. Happy to do either way, the server just won’t work properly without one. Should we default to stdio if it’s missing (typo maybe in the config?) or just return an error. So far I’m leaning towards the error.

internal/server/server.go
keremgocen: I will double check but I think it’s already included in the SDK path once we use the MCP server handler as the base. Which allows us to add more endpoints like /health at the root level for CI.

internal/server/server.go
keremgocen: The built-in server start function works just fine, I’ve been playing around with the setup for a bit. We said we will have some health and dbinfo functionality here but I think we also said they will be tool calls within the MCP server. If we do so I think we can use the built in http server, but if we want a custom multiplexer we have to use this way.
So the question is do we want additional endpoints such as `/health` and `/dbinfo` or are we happy to keep them inside as tool calls?

internal/cli/args.go
MacondoExpress: I was meant to keep it only the argument but not the environment variable, this thought was driven by the desire the reduce the documentation/testing surface of the solution, but after thinking the night on this, I think that in this case having "parity" between envs and arguments is a good choice here. 
What makes me leaning to the current implementation is that even if the transport mode is unlikely to be changed depending on the host or environment at the deploy time, all the closed related settings are!

internal/server/server.go
MacondoExpress: > I guess this makes it a required config right? 

No, in the way in which I thought this, we have a default one (STDIO), I would say!

> Should we default to stdio if it’s missing (typo maybe in the config?) or just return an error.

Strongly leaning on this behaviour: if it's set, either by environment variable or arguments, then we validate it and raise the error in case of wrong input, if not we default to STDIO.
To me in this way, we have the best of the two world, expected behaviour, best ergonomic!

cmd/neo4j-mcp/main.go
MacondoExpress: Up to you, I wanted to highlight this thought, but I'm happy either way!

internal/server/server.go
MacondoExpress: I would personally start with the more straightforward solution first, then think about the `/health`, `/dbinfo` endpoints, when the need arises (I'm not 100% it will!).

keremgocen: ok, I think we'll use the built-in start and not have any custom endpoints for now. This might change later in Aura.

internal/server/server.go
keremgocen: We'll default to STDIO and raise a validation error if we can't map the input to one of the valid transport options (enum-like).

cmd/neo4j-mcp/main.go
keremgocen: I'll try to clean up main a bit

internal/cli/args.go
```diff
@@ -25,6 +26,7 @@ Options:
   --neo4j-read-only <BOOLEAN>         Enable read-only mode: true or false (overrides environment variable NEO4J_READ_ONLY)
   --neo4j-telemetry <BOOLEAN>         Enable telemetry: true or false (overrides environment variable NEO4J_TELEMETRY)
   --neo4j-schema-sample-size <INT>    Number of nodes to sample for schema inference (overrides environment variable NEO4J_SCHEMA_SAMPLE_SIZE)
+  --neo4j-transport-mode <MODE>       MCP Transport mode (e.g., 'stdio', 'http') (overrides environment variable NEO4J_MCP_TRANSPORT)
```
MacondoExpress, line 29: Missing port and host as arguments

test/integration/dbservice/dbservice.go
```diff
@@ -62,15 +62,21 @@ func (dbs *dbService) GetDriver() *neo4j.DriverWithContext {
 	return dbs.driver
 }
 
-func (dbs *dbService) GetDriverConf() *config.Config {
+func (dbs *dbService) GetDriverConf(transport string) *config.Config {
```
MacondoExpress: I would remove the ability to specify here the transport, it creates not obviously behaviours.
```suggestion
func (dbs *dbService) GetDriverConf() *config.Config {
```

internal/cli/args.go
keremgocen: added them as flags

PR #117, opened 2025-11-28
Remove "params" from required when responding to "tools/list" call

MacondoExpress:
Remove "params" argument from required fields for Cypher tools (read-cypher/write-cypher).

Made the `params` field optional in both `ReadCypherInput` and `WriteCypherInput` structs by adding the `omitempty` JSON tag


LackOfMorals (commented):
Let

PR #120, opened 2025-12-01
Basic auth and allowed origins for read-cypher tool

keremgocen:
Introduce transport mode configuration for the server and implement a basic HTTP server. Enhance tests to include transport mode and add authentication context and middleware for secure query execution. Clean up logging and fix minor issues in the codebase.

All tools updated, the service layer is now using an auth helper which extracts the credentials from context per request for the HTTP mode.
The driver is reused via impersonation https://neo4j.com/docs/go-manual/current/query-simple/#impersonation instead of creating a new one per request or tool call, since the connection string does not change.

### DIFF
--- a/internal/tools/types.go
+++ b/internal/tools/types.go
@@ -2,7 +2,6 @@ package tools
 
 import (
 	"github.com/neo4j/mcp/internal/analytics"
-	"github.com/neo4j/mcp/internal/config"
 	"github.com/neo4j/mcp/internal/database"
 )
 
@@ -11,5 +10,4 @@ type ToolDependencies struct {
 	DBService        database.Service
 	AnalyticsService analytics.Service
 	SchemaSampleSize int
-	Config           *config.Config
 }
--- a/internal/server/tools_register.go
+++ b/internal/server/tools_register.go
@@ -48,7 +48,6 @@ func (s *Neo4jMCPServer) getEnabledTools() []server.ServerTool {
 	deps := &tools.ToolDependencies{
 		DBService:        s.dbService,
 		AnalyticsService: s.anService,
-		Config:           s.config,
 	}
 	toolDefs := s.getAllToolsDefs(deps)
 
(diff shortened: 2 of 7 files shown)


internal/config/config.go
```diff
@@ -67,6 +58,17 @@ func (c *Config) Validate() error {
 		return fmt.Errorf("invalid transport mode '%s', must be one of %v", c.TransportMode, ValidTransportModes)
 	}
 
+	// For STDIO mode, require username and password from environment
+	// For HTTP mode, credentials come from per-request Basic Auth headers
```
MacondoExpress, line 62: I would raise a validation error here for cases when users defines the username and password at startup in http mode, so that the behaviour is always expected and there is no confusion that as example NEO4J_USERNAME will be no used in http mode.

internal/server/middleware.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/neo4j/mcp/internal/auth"
+)
+
+const (
+	maxAgeSeconds = "86400" // 24 hours
+)
+
+// addMiddleware chains together all HTTP middleware
+func addMiddleware(allowedOrigins []string, next http.Handler) http.Handler {
```
MacondoExpress: Suggestion: The name addMiddleware implies a generic utility for appending any middleware to a chain (an agnostic helper). However, this function strictly defines our specific HTTP pipeline (CORS, Auth, Logging).
I would suggest to change the name to reflect this!

internal/tools/types.go
```diff
@@ -10,4 +11,5 @@ type ToolDependencies struct {
 	DBService        database.Service
 	AnalyticsService analytics.Service
 	SchemaSampleSize int
+	Config           *config.Config
```
MacondoExpress: Config is no longer a ToolDependency needed

internal/server/middleware.go
keremgocen: I was thinking about `addMiddleware(s)` or `applyMiddleware` instead but I think this is fine, for lack of a better function name, which is usually hard.

I don't want to restrict the name to `auth` either as they are not purely auth. We have logging, CORS and might have more in the future.

How about `chainMiddleware` if we must change it? (I think not)

internal/tools/types.go
keremgocen: Good catch, that cleans up some test code as well

PR #123, opened 2025-12-02
Add Contributing guide on how to build an MCPB bundle.

MacondoExpress:
As title, 
this PR add the contributing guide to set the way on how to build an `.mcpb` bundle: https://github.com/modelcontextprotocol/mcpb/

Due to a limitation over the ability to select the binary depending on the host architecture we opted to not release the bundle publicly yet.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -216,34 +216,11 @@ When adding new database operations:
 - Tests failing unexpectedly → regenerate mocks, verify env vars, rerun full test suite.
 - Dependency/build issues → `go mod tidy`.
 
-## Build MCPB Bundle (for Claude Desktop)
+## Update MCPB Bundle (for Claude Desktop)
 
-You can package the MCP server into an `.mcpb` bundle for distribution with MCP clients (e.g., Anthropic/Claude).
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -214,6 +216,35 @@ When adding new database operations:
 - Tests failing unexpectedly → regenerate mocks, verify env vars, rerun full test suite.
 - Dependency/build issues → `go mod tidy`.
 
+## Build MCPB Bundle (for Claude Desktop)
```
keremgocen: Should we move the client-specific config into their own separate files? We have a lot of clients, and I think this is irrelevant for anyone not using claude desktop, bloating the contributing file unnecessarily. 

The mcpb setup maybe even calls for it's own doc page also stating what an .mcpb bundle is?

We should also mention that manifest.json needs to be updated when there's a new env variable.

MacondoExpress: Good points, was actually thinking about this when I added, maybe I left a small link to here for visibility but the content in another document!

MacondoExpress: Updated following this comment.

keremgocen (approved):
LGTM

PR #126, opened 2025-12-11
Add TLS support for secure communication

keremgocen:
Adds TLS support for the Neo4j MCP server, with path validation and security improvements.

⚠️ Unit tests are using a pre-generated cert via the helper function.

Summary of changes:
- Add pathValidationMiddleware() to reject non-/mcp paths with fast 404 response
- Prevents hanging connections on invalid endpoints
- Document certificate CN/SAN requirements for domain verification (new guide: TLS_SETUP.md)
- Add examples for localhost vs production certificate generation
- Update timeout values for better security/performance balance:
  - ReadHeaderTimeout: 5s (prevents Slowloris attacks)
  - ReadTimeout: 10s → 15s (prevents slow-read attacks)
  - IdleTimeout: 60s → 120s (improves connection reuse)
- Add clear security/functionality comments for each timeout
- Use InsecureSkipVerify for self-signed test certificates (normally triggers a linter warning, but ignored for tests)

### DIFF
--- a/internal/cli/args.go
+++ b/internal/cli/args.go
@@ -44,7 +44,7 @@ Optional Environment Variables:
   NEO4J_READ_ONLY Enable read-only mode (default: false)
   NEO4J_SCHEMA_SAMPLE_SIZE Number of nodes to sample for schema inference (default: 100)
   NEO4J_MCP_TRANSPORT MCP Transport mode (e.g., 'stdio', 'http') (default: stdio)
-  NEO4J_MCP_HTTP_PORT HTTP server port (default: 443 with TLS, 8080 without TLS)
+  NEO4J_MCP_HTTP_PORT HTTP server port (default: 443 with TLS, 80 without TLS)
   NEO4J_MCP_HTTP_HOST HTTP server host (default: 127.0.0.1)
   NEO4J_MCP_HTTP_TLS_ENABLED Enable TLS/HTTPS for HTTP server (default: false)
   NEO4J_MCP_HTTP_TLS_CERT_FILE Path to TLS certificate file (required when TLS is enabled)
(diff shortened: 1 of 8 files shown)


MacondoExpress (approved):
LGTM, 
I leave here some feedbacks:
- Could be very beneficial for the administrator to receive warnings when the certificate is close to be expired.
- Currently seems only x509 PEM encoded certificate are supported, which is completely fine to me, however it will be good to be explicit on the capability of the server and handle when user pass unsupported certificates.
- I've tested with the inspector and with VSCode and it was failing cause the self signed certificate, I don't think it's an issue with your implementation however, it will be good to document for the beneficial of both end users and maintainers (the error is not very intuitive)


I have other comments regarding the documentation but I'll add them directly on the main PR as they are not part of this PR.
Maybe worth give the time also to @mjfwebb  and @Liam-Doodson to look at this

internal/cli/args.go
```diff
@@ -41,8 +44,11 @@ Optional Environment Variables:
   NEO4J_READ_ONLY Enable read-only mode (default: false)
   NEO4J_SCHEMA_SAMPLE_SIZE Number of nodes to sample for schema inference (default: 100)
   NEO4J_MCP_TRANSPORT MCP Transport mode (e.g., 'stdio', 'http') (default: stdio)
-  NEO4J_MCP_HTTP_PORT HTTP server port (default: 8080)
+  NEO4J_MCP_HTTP_PORT HTTP server port (default: 443 with TLS, 8080 without TLS)
```
Liam-Doodson: As a minor nit - why `443` and `8080`? Surely, it would make more sense to either default to `443` & `80` or `8443` and `8080`?

keremgocen: Good point, because I started testing with 8080 initially then I said it’s TLS so the default should be 443. I think 443 & 80 makes more sense here since we have no other config or proxy around it. I’ll update the PR in the afternoon before merging.

PR #129, opened 2025-12-16
Add End-to-End Test Suite for MCP Server

MacondoExpress:
This PR introduces a end-to-end (E2E) testing framework for the Neo4j MCP server, providing full server lifecycle testing including compilation, MCP protocol communication, and tool execution.

How E2E Testing Works:
* Binary compilation and reuse - Server binary is compiled once in TestMain and stored in a temporary directory, then reused across all test cases for efficiency
* Real MCP protocol communication - Each test creates a fresh StdioMCPClient that communicates with the compiled server binary via stdin/stdout, simulating actual MCP usage

GH Action workflow has changed to includ the e2e tests coverage.

Note:
This initial coverage demonstrates the test suite capabilities and infrastructure. Future PRs should expand coverage to include additional MCP tools, error scenarios, and edge cases to ensure comprehensive validation of the server functionality.

### DIFF
--- a/test/e2e/get_schema_test.go
+++ b/test/e2e/get_schema_test.go
@@ -10,7 +10,6 @@ import (
 	"github.com/mark3labs/mcp-go/client"
 	"github.com/mark3labs/mcp-go/mcp"
 	"github.com/neo4j/mcp/test/e2e/helpers"
-	"github.com/stretchr/testify/assert"
 )
 
 func TestGetSchemaE2E(t *testing.T) {
@@ -38,39 +37,6 @@ func TestGetSchemaE2E(t *testing.T) {
 	}
 	t.Cleanup(func() {
 		mcpClient.Close()
-	})
-	t.Run("get-schema with empty database", func(t *testing.T) {
-		t.Parallel()
-		// Call get-schema tool on empty database
-		callToolRequest := mcp.CallToolRequest{
-			Params: mcp.CallToolParams{
-				Name: "get-schema",
-			},
-		}
-
-		callToolResponse, err := mcpClient.CallTool(ctx, callToolRequest)
-		if err != nil {
-			t.Fatalf("failed to call get-schema tool: %v", err)
-		}
-		textContent, ok := mcp.AsTextContent(callToolResponse.Content[0])
-		if !ok {
-			t.Fatalf("expected error as TextContent, got %T", callToolResponse.Content[0])
-		}
-		// Verify the tool call was successful
-		if callToolResponse.IsError {
-			t.Fatalf("get-schema tool call returned an error: %s", textContent.Text)
-		}
-
-		// For empty database, should return a message indicating no schema
-		if len(callToolResponse.Content) == 0 {
-			t.Fatal("expected get-schema tool to return content, but got none")
-		}
-
-		// Should contain message about empty database
-		expectedMessage := "The get-schema tool executed successfully; however, since the Neo4j instance contains no data, no schema information was returned."
-		assert.Equal(t, expectedMessage, textContent.Text, "Should return empty database message")
-
-		t.Log("Successfully handled get-schema on empty database")
 	})
 
 	t.Run("get-schema with nodes only", func(t *testing.T) {


test/e2e/main_test.go
```diff
@@ -0,0 +1,36 @@
...
+	if err != nil {
+		log.Fatal("error while creating MCP server for e2e purpose")
+	}
+	dbs.Start(ctx)
+
+	code := m.Run()
+
+	dbs.Stop(ctx)
+
+	cleanUpServerDir()
```
keremgocen, line 33: We should use defer here in case something panics in tests and cleanup fails
```suggestion
	defer cleanUpServerDir()
```

test/containerrunner/container_runner.go
```diff
@@ -1,4 +1,4 @@
-//go:build integration
+//go:build integration || e2e
```
keremgocen, line 1: I did not know you could combine multiple build tags like this, nice! (assuming this will be built in both cases when we use either of these tags)

test/e2e/helpers/build_server.go
```diff
@@ -0,0 +1,66 @@
...
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+)
+
+// BuildServer compile the project and store the output binary in a TempDir,
+// Returns a callback to delete the temporary directory when is is no longer needed.
+func BuildServer() (string, func(), error) {
```
keremgocen, line 15: Do we need to build the binary every time? A lot of manual steps here sounds a bit brittle to maintain long-term to me. Could we use the Dockerfile instead, for example?

test/e2e/main_test.go
```diff
@@ -0,0 +1,36 @@
...
+func TestMain(m *testing.M) {
+	ctx := context.Background()
+
+	srv, cleanUpServerDir, err := helpers.BuildServer()
+	server = srv
+
+	if err != nil {
+		log.Fatal("error while creating MCP server for e2e purpose")
+	}
+	dbs.Start(ctx)
```
keremgocen, line 27: this is a common issue with integration tests but I'm wondering if we need to add some logic to make the tests wait until the database is ready, or if containerrunner already waits for an available port?

test/e2e/mcp_server_lifecycle_test.go
```diff
@@ -0,0 +1,86 @@
...
+			if !ok {
+				t.Fatalf("expected error as TextContent, got %T", callToolResponse.Content[0])
+			}
+			t.Fatalf("get-schema tool call returned an error: %s", textContent.Text)
+		}
+
+		if len(callToolResponse.Content) == 0 {
+			t.Fatal("expected get-schema tool to return content, but got none")
+		}
+		mcpClient.Close()
```
keremgocen: most of these are wrapped in `t.Cleanup` or deferred, I guess this one is missed?

test/e2e/mcp_server_lifecycle_test.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/mark3labs/mcp-go/client"
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/neo4j/mcp/test/e2e/helpers"
+)
+
+func TestSeverLifecycleMCP(t *testing.T) {
```
keremgocen: Should we keep the e2e test names consistent (suggestion: either make all of them end with E2E or alternatively, we can rely on the build tag only and name them as usual)

test/e2e/helpers/helpers.go
```diff
@@ -0,0 +1,169 @@
...
+	TestID        string
+	Service       database.Service
+	createdLabels map[string]bool
+}
+
+// NewE2ETestContext creates a new E2E test context with automatic cleanup
+func NewE2ETestContext(t *testing.T, driver *neo4j.DriverWithContext) *E2ETestContext {
+	t.Helper()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
keremgocen: this seems a bit too long, do we need 5 minutes timeout for the entire e2e suite? could this potentially hide other issues?

keremgocen (approved):
This looks good to me for a start, but ideally, we should consider adding a common interface for test setup for both transport modes in the near future (and maybe document now that these are only working for STDIO mode). Maybe we can also check that read-only mode does not allow write queries and gradually expand these.

test/e2e/helpers/build_server.go
MacondoExpress: It will build only once for all the tests in the e2e suite, the scope is to have a test that tests that is closed to the real world scenario. If we test them in a container we loose some of the benefits like testing in a specific environment.

test/e2e/helpers/helpers.go
MacondoExpress: No it's true, it's too high we can reduce it!

test/e2e/main_test.go
MacondoExpress: It will be no executed, it's followed by `os.Exit`

test/e2e/main_test.go
MacondoExpress: No, when `dbs.Start` returns the container is already available and listening!

PR #134, opened 2025-12-17
Add Bearer token authentication middleware

keremgocen:
Introduce middleware to support Bearer token authentication alongside Basic Auth. This enhancement allows for more flexible authentication options, particularly for SSO/OAuth scenarios. The implementation includes context management for Bearer tokens and corresponding tests to ensure functionality.

- [x] updates auth package to store Bearer tokens in context
- [x] test coverage for database service

### DIFF
--- a/internal/database/service_auth_config_test.go
+++ b/internal/database/service_auth_config_test.go
@@ -43,10 +43,6 @@ func TestBuildQueryOptions_HTTPMode_BearerToken(t *testing.T) {
 	if config.Auth == nil {
 		t.Fatal("Expected auth token to be set, got nil")
 	}
-
-	// Verify it's a bearer token (BearerAuth sets a specific scheme)
-	// The neo4j driver stores tokens with their scheme, we can't inspect the exact token
-	// but we can verify that Auth was populated
 }
 
 // TestBuildQueryOptions_HTTPMode_BasicAuth verifies that basic auth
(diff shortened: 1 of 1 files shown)


internal/auth/auth_test.go
```diff
@@ -0,0 +1,107 @@
...
+	if token != "" {
+		t.Errorf("Expected empty token when no bearer token, got %q", token)
+	}
+}
+
+// Note: We don't test empty bearer tokens because the middleware (authMiddleware)
+// explicitly rejects empty tokens with a 401 error before they reach the context.
+// See internal/server/middleware.go, an empty bearer token can never exist in context in production.
+
+func TestBothBearerAndBasicAuthInContext(t *testing.T) {
```
keremgocen: while this test looks correct, I don't think we should get multiple authorization headers within the same request so I think there's no need to worry about the auth precedence. As it seems to be a syntax error while some servers apparently do accept it. https://stackoverflow.com/a/38515091

internal/server/middleware.go
```diff
@@ -35,21 +36,51 @@ func chainMiddleware(allowedOrigins []string, next http.Handler) http.Handler {
...
+				next.ServeHTTP(w, r.WithContext(ctx))
+				return
+			}
+
+			// Fall back to basic auth
 			user, pass, ok := r.BasicAuth()
 			if !ok {
 				// No credentials provided - reject request
+				w.Header().Add("WWW-Authenticate", `Basic realm="Neo4j MCP Server"`)
+				w.Header().Add("WWW-Authenticate", `Bearer realm="Neo4j MCP Server"`)
```
MacondoExpress, line 71:
```suggestion
```

keremgocen: Are you sure we want to remove this? This is used to advertise which auth methods the server is supporting. In fact, I'm thinking if we should advertise both headers ("WWW-Authenticate") in both cases (basic+bearer)?

MacondoExpress: Thank you for questioning this, I didn't see the difference between the two lines at first glance and I thought it was a typo!

PR #135, opened 2025-12-18
Add HTTP mode metrics collection and improvements to metrics

keremgocen:
HTTP Mode Telemetry Support:
- Added telemetry collection for HTTP transport mode
- Emits startup metrics (transport mode, TLS status) on first HTTP request
- Connection initialised event sent on first tool call (HTTP) vs startup (STDIO)                                                                   
- Separated `MCP_STARTUP` and `CONNECTION_INITIALIZED` events as the latter requires a valid DB connection to obtain the information, which happens only after the first successful request for HTTP

Telemetry improvements:
- All tool events track success/failure for reliability metrics.
- Added `mcpVersion` field to analytics configuration.
- Added MCP SDK Hooks Integration, which uses the SDK's hook mechanism to send events after the tool call is complete. This also helped clean up manual event emissions from the handlers.
- Removed `httpMetricsMiddleware()` method entirely.

⚠️ Breaking Changes:
- Startup event no longer includes DB info and is shared for both transport modes.
- A new connection initialisation event is sent after the initial database connection. (immediately for STDIO, after the first connected request for HTTP mode)
- All tool events now include a `success` boolean field. (also shared for both modes)

Documentation
- Added warning about Neo4j 5.26.18 APOC bug that breaks get-schema tool (fixed in 5.26.19+)

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -42,7 +42,7 @@ jobs:
         run: go test -v -race ./...
 
   integration-tests:
-    name: Build and Integration Tests
+    name: Integration Tests
     strategy:
       fail-fast: false
       matrix:
@@ -65,6 +65,36 @@ jobs:
 
       - name: Run integration tests
         run: go test -v -tags=integration ./test/integration/...
+        env:
+          USE_CONTAINER: "false"
+          NEO4J_URI: ${{ secrets.AURA_URL }}
+          NEO4J_USERNAME: ${{ secrets.AURA_USERNAME }}
+          NEO4J_PASSWORD: ${{ secrets.AURA_PASSWORD }}
+
+  e2e-tests:
+    name: E2E Tests
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
+
+      - name: Set up Go
+        uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c # v6
+        with:
+          go-version: "1.25.4"
+
+      - name: Download dependencies
+        run: go mod download
+
+      - name: Build
+        run: go build -v ./...
+
+      - name: Run E2E tests
+        run: go test -v -tags=e2e ./test/e2e/...
         env:
           USE_CONTAINER: "false"
           NEO4J_URI: ${{ secrets.AURA_URL }}
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -33,6 +33,9 @@ tasks:
   test:int:
     cmds:
       - go test -tags=integration ./test/integration/... {{.CLI_ARGS}}
+  test:e2e:
+    cmds:
+      - go test -tags=e2e ./test/e2e/... {{.CLI_ARGS}}
   test:clean:
     cmds:
       - go clean -testcache
@@ -42,3 +45,6 @@ tasks:
   inspect:
     cmds:
       - npx @modelcontextprotocol/inspector go run ./cmd/neo4j-mcp
+  interactive-client:
+    cmds:
+      - go run ./client/... bin/neo4j-mcp
--- a/test/containerrunner/container_runner.go
+++ b/test/containerrunner/container_runner.go
@@ -1,4 +1,4 @@
-//go:build integration
+//go:build integration || e2e
 
 package containerrunner
 
--- a/test/dbservice/dbservice.go
+++ b/test/dbservice/dbservice.go
@@ -1,4 +1,4 @@
-//go:build integration
+//go:build integration || e2e
 
 package dbservice
 
@@ -8,7 +8,7 @@ import (
 	"sync"
 
 	"github.com/neo4j/mcp/internal/config"
-	"github.com/neo4j/mcp/test/integration/containerrunner"
+	"github.com/neo4j/mcp/test/containerrunner"
 	"github.com/neo4j/neo4j-go-driver/v6/neo4j"
 )
 
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -1,0 +1,170 @@
+# End-to-End (E2E) Tests
+
+End-to-end tests for the Neo4j MCP server that test the complete server lifecycle including compilation, MCP protocol communication, and tool execution using a shared Neo4j container (includes APOC + GDS).
+
+## Quick Start
+
+```go
+func TestMyE2EFeature(t *testing.T) {
+    t.Parallel()
+
+    // Create MCP client
+    ctx := context.Background()
+    cfg := dbs.GetDriverConf()
+    args := []string{
+        "--neo4j-uri", cfg.URI,
+        "--neo4j-username", cfg.Username,
+        "--neo4j-password", cfg.Password,
+        "--neo4j-database", cfg.Database,
+    }
+
+    mcpClient, err := client.NewStdioMCPClient(server, []string{}, args...)
+    if err != nil {
+        t.Fatalf("failed to create MCP client: %v", err)
+    }
+
+    // Initialize the server
+    _, err = mcpClient.Initialize(ctx, helpers.BuildInitializeRequest())
+    if err != nil {
+        t.Fatalf("failed to initialize MCP server: %v", err)
+    }
+    t.Cleanup(func() {
+        mcpClient.Close()
+    })
+
+    // Create test context for database operations (auto-isolation + cleanup)
+    tc := helpers.NewE2ETestContext(t, dbs.GetDriver())
+
+    // Seed test data (automatically isolated with unique labels and cleaned up)
+    personLabel, err := tc.SeedNode("Person", map[string]any{"name": "Alice"})
+    if err != nil {
+        t.Fatalf("failed to seed data: %v", err)
+    }
+
+    // Call MCP tool
+    callToolRequest := mcp.CallToolRequest{
+        Params: mcp.CallToolParams{
+            Name: "read-cypher",
+            Arguments: map[string]any{
+                "query":  "MATCH (p:" + personLabel.String() + " {name: $name}) RETURN p",
+                "params": map[string]any{"name": "Alice"},
+            },
+        },
+    }
+
+    callToolResponse, err := mcpClient.CallTool(ctx, callToolRequest)
+    if err != nil {
+        t.Fatalf("failed to call tool: %v", err)
+    }
+
+    // Verify response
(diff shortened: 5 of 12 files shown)


internal/server/middleware.go
```diff
@@ -1,28 +1,34 @@
...
 )
 
 const (
 	corsMaxAgeSeconds = "86400" // 24 hours
 )
 
-// chainMiddleware chains together all HTTP middleware
-func chainMiddleware(allowedOrigins []string, next http.Handler) http.Handler {
+// chainMiddleware chains together all HTTP middleware for this server instance
+func (s *Neo4jMCPServer) chainMiddleware(allowedOrigins []string, next http.Handler) http.Handler {
```
keremgocen, line 17: we could move this into the server.go potentially, still in the same package so for now I'm keeping it since it's part of the middleware logic

internal/server/middleware.go
```diff
@@ -130,3 +136,28 @@ func loggingMiddleware() func(http.Handler) http.Handler {
...
+// httpMetricsMiddleware collects and emits HTTP mode metrics on the first request.
+// Uses sync.Once to ensure metrics are collected exactly once per server session.
+// Requires Basic Auth credentials to be present in the request context (set by basicAuthMiddleware).
+func (s *Neo4jMCPServer) httpMetricsMiddleware() func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Only collect metrics if telemetry is enabled AND transport mode is HTTP
+			if s.config.Telemetry && s.config.TransportMode == config.TransportModeHTTP {
+				// Use sync.Once to ensure metrics are collected exactly once
+				s.httpMetricsSent.Do(func() {
```
keremgocen: this should only be called once per server run, added a unit test to ensure that behaviour

internal/server/middleware.go
```diff
@@ -130,3 +136,28 @@ func loggingMiddleware() func(http.Handler) http.Handler {
 		})
 	}
 }
+
+// httpMetricsMiddleware collects and emits HTTP mode metrics on the first request.
+// Uses sync.Once to ensure metrics are collected exactly once per server session.
+// Requires Basic Auth credentials to be present in the request context (set by basicAuthMiddleware).
```
Liam-Doodson: Do we need to check the auth requirement before sending the metrics? Just trying to think what happens if the header is missing/invalid in the first request?

keremgocen: good catch, I think because we passing a new context to the metrics collection we'll lose the auth in that one and the query will fail :(
```
go func() {
  ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second) // auth credentials are in r.Context() actually
  defer cancel()
  s.collectAndEmitHTTPMetrics(ctx)
}()
```

internal/server/middleware_test.go
```diff
@@ -423,3 +462,132 @@ func TestPathValidationMiddleware_InFullChain(t *testing.T) {
...
+					// After each request, wait for any spawned goroutines to complete
+					// This ensures we're testing that sync.Once prevents multiple collections
+					// across sequential requests, not just within a single request
+					if tt.expectMetrics {
+						synctest.Wait()
+					}
+				}
+
+				// Final verification: mock expectations with .Times(1) will fail if
+				// metrics were collected more than once across all 3 requests
```
Liam-Doodson: Is this still WIP?

internal/analytics/events.go
```diff
@@ -65,15 +71,29 @@ type StartupEventInfo struct {
...
+		baseProperties: a.getBaseProperties(),
+		Neo4jVersion:   startupInfoEvent.Neo4jVersion,
+		Edition:        startupInfoEvent.Edition,
+		CypherVersion:  startupInfoEvent.CypherVersion,
+		McpVersion:     startupInfoEvent.McpVersion,
+		TransportMode:  a.cfg.transportMode,
+	}
+
+	// For HTTP mode, include TLS-specific properties
+	var properties any
```
MacondoExpress: I personally would find the implementation cleaner if we just include `tls_enabled` as a property of startupProperties, without the need to create a specialised StartupEvent for HTTP, it can be set "false" for STDIO startup events.

internal/server/middleware.go
```diff
@@ -130,3 +136,33 @@ func loggingMiddleware() func(http.Handler) http.Handler {
...
 }
+
+// httpMetricsMiddleware collects and emits HTTP mode metrics on the first request.
+// Uses sync.Once to ensure metrics are collected exactly once per server session.
+// Extracts Basic Auth credentials from request context for Neo4j query authentication.
+func (s *Neo4jMCPServer) httpMetricsMiddleware() func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Only collect metrics if telemetry is enabled AND transport mode is HTTP
+			if s.config.Telemetry && s.config.TransportMode == config.TransportModeHTTP {
```
MacondoExpress:
```suggestion
			if s.config.Telemetry  {
```
I would assume that we are on HTTP mode inside an http.Handler!

MacondoExpress (approved):

internal/server/middleware.go
keremgocen: Looks like I was being extra safe:)

internal/server/middleware.go
keremgocen: we made a bunch of changes to this logic but basically `sync.Once` won't be consumed anymore unless there's valid auth credentials in the context, this was the initial bug

internal/analytics/events.go
keremgocen: we are using the same container now for both modes, `omitempty` is used for this field so we drop it when empty in STDIO mode

internal/server/server.go
```diff
@@ -380,3 +423,76 @@ func (s *Neo4jMCPServer) StartHTTPServer() error {
...
+	// HTTP mode only: emit connection initialized event on first successful tool call
+	// (STDIO mode emits this at startup instead)
+	if s.config.TransportMode == config.TransportModeHTTP {
+		// Collect connection info once for CONNECTION_INITIALIZED event
+		if s.connectionInfo.Neo4jVersion == "" {
+			connInfo := s.collectConnectionInfo(ctx)
+			// Only cache if we got valid connection info (not empty, not "unknown")
+			if connInfo.Neo4jVersion != "" && connInfo.Neo4jVersion != "unknown" {
+				s.connectionInfo = connInfo // Cache for future tool calls
+			}
```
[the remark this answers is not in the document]
keremgocen: We don't expect this field to be modified after start, but it's still a valid suggestion to have a mutex on it before accessing. I'll add this.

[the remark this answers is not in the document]
keremgocen: done, I'm still thinking about if we can simplify this logic though... but we now have a mutex to protect connInfo for read/wrtes

.changes/unreleased/Patch-20260114-132430.yaml
```diff
@@ -0,0 +1,5 @@
+kind: Patch
+body: |
+  Added HTTP mode telemetry that collects and emits startup metrics (transport mode, TLS status, Neo4j version) on the first HTTP request when telemetry is enabled. Includes comprehensive test coverage with proper goroutine synchronization using testing/synctest.
```
MacondoExpress:
```suggestion
```
I would keep the bearer token support for MCP server, but I would consider a `minor` as new functionality.
Around the analytics and test coverage details I'll consider them implementation details that do not requires any kind intervention by the end user.
If we want to keep an analytic patch note, I would keep communicate only what's in the user interest, something like: `Analytics events are now sent also when the server is executed in HTTP mode; previously, only STDIO mode supported Telemetry.`

internal/analytics/analytics.go
```diff
@@ -31,7 +34,7 @@ type Analytics struct {
 }
 
 // for testing purposes - enables dependency injection of http client
-func NewAnalyticsWithClient(mixPanelToken string, mixpanelEndpoint string, client HTTPClient, uri string) *Analytics {
+func NewAnalyticsWithClient(mixPanelToken string, mixpanelEndpoint string, client HTTPClient, uri string, transportMode string, tlsEnabled bool, mcpVersion string) *Analytics {
```
MacondoExpress: Let's transport `transportMode`, `tlsEnabled`, and `mcpVersion` directly into the constructor of the event that needs them, in this case, only the Startup one. 
This will avoid having a parameter expansion as the events grow in population and/or in depth.

internal/analytics/analytics.go
```diff
@@ -40,12 +43,15 @@ func NewAnalyticsWithClient(mixPanelToken string, mixpanelEndpoint string, clien
...
+		transportMode:    transportMode,
+		tlsEnabled:       tlsEnabled,
+		mcpVersion:       mcpVersion,
 	}
 
 	return &Analytics{cfg: cfg, disabled: false}
 }
 
-func NewAnalytics(mixPanelToken string, mixpanelEndpoint string, uri string) *Analytics {
+func NewAnalytics(mixPanelToken string, mixpanelEndpoint string, uri string, transportMode string, tlsEnabled bool, mcpVersion string) *Analytics {
```
MacondoExpress: same as above!

internal/server/server.go
```diff
@@ -380,3 +426,88 @@ func (s *Neo4jMCPServer) StartHTTPServer() error {
...
+	if s.anService == nil || !s.anService.IsEnabled() {
+		return
+	}
+
+	toolName := request.Params.Name
+	success := !result.IsError
+
+	// HTTP mode only: emit connection initialized event on first successful tool call
+	// (STDIO mode emits this at startup instead)
+	if s.config.TransportMode == config.TransportModeHTTP {
```
MacondoExpress: maybe taking the changes from https://github.com/neo4j/mcp/pull/156 as this logic will be handled for for free

MacondoExpress (changes requested):
Looks good, the hook "handleToolCallComplete" really simplify the analytics ergonomic with tool calling!
Left some comments, the main one are:
- the analyticService constructor, I think that adding event information there will be no scalable/correct.
- the way how connectionInfo is collected, we can take the changes from https://github.com/neo4j/mcp/pull/156 and it will removes the extra complexity

internal/analytics/analytics.go
```diff
@@ -114,7 +118,7 @@ func (a *Analytics) sendTrackEvent(events []TrackEvent) error {
 	return nil
 }
 
-func getDistinctID() string {
+func GetDistinctID() string {
```
MacondoExpress, line 121: Not-blocking: GetDistinct is not used outside Analytics

keremgocen: we skipped it during pairing but my plan was to move the `NewAnalyticsWithClient` to test package since it's only used in tests, then it would have to call `analytics. GetDistinctID`. I'll do that in a later PR.

PR #153, opened 2026-01-13
Add txMetadata app name in Neo4jMCP queries

MacondoExpress:
As title, this PR includes txMetadata "app" field set as "NEO4JMCP/<version>" to identify Neo4j MCP as origin of Neo4j queries.

### DIFF
--- a/internal/database/service.go
+++ b/internal/database/service.go
@@ -12,7 +12,7 @@ import (
 	"github.com/neo4j/neo4j-go-driver/v5/neo4j"
 )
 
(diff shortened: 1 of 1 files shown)


internal/database/service.go
```diff
@@ -12,23 +12,27 @@ import (
 	"github.com/neo4j/neo4j-go-driver/v5/neo4j"
 )
 
+const APP_NAME string = "MCP4NEO4J"
```
keremgocen: yes please, we can use `appName`, which is perfectly fine and consistent with the rest of our constants.

PR #155, opened 2026-01-15
Add lazy initialization and atomic synchronization for driver and connection info

keremgocen:
Implement lazy initialization for the driver and replace mutex with atomic value for connection info synchronization.

 Benefits:                                                                                                              
 - Simpler code - no manual lock/unlock                                                                                 
- Better performance and lock-free operations                                                                            
- Safer, impossible to forget to unlock                                                                               
- Natural fit for write-once scenarios

### DIFF
--- a/test/integration/dbservice/dbservice.go
+++ b/test/integration/dbservice/dbservice.go
@@ -14,7 +14,7 @@ import (
 
 type dbService struct {
 	driver       *neo4j.Driver
-	driverMu     sync.Mutex // Protects driver from concurrent access
+	driverOnce   sync.Once // Ensures driver is initialized exactly once
 	useContainer bool
 }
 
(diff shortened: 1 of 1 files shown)


test/integration/dbservice/dbservice.go
```diff
@@ -37,7 +39,10 @@ func (dbs *dbService) Stop(ctx context.Context) {
 	}
 }
 
-func (dbs *dbService) GetDriver() *neo4j.DriverWithContext {
+func (dbs *dbService) GetDriver() *neo4j.Driver {
+	dbs.driverMu.Lock()
```
keremgocen: why we need the mutex here? Without it, multiple parallel tests would simultaneously detect `driver == nil`, and all try to create and assign drivers, causing the race condition in the test output.

internal/server/server.go
```diff
@@ -44,9 +45,8 @@ type Neo4jMCPServer struct {
 	version                string
 	anService              analytics.Service
 	gdsInstalled           bool
-	connectionInfo         analytics.ConnectionEventInfo // Cached connection info (STDIO: set at startup, HTTP: cached after first tool call)
-	connectionInfoMu       sync.RWMutex                  // Protects connectionInfo from concurrent access
-	httpConnectionInitSent sync.Once                     // Ensures connection initialized event is sent only once in HTTP mode
+	connectionInfo         atomic.Value // stores analytics.ConnectionEventInfo; cached connection info (STDIO: set at startup, HTTP: cached after first tool call)
```
keremgocen, line 48: I don't recall using sync/atomic a lot in the past, but this felt like a better alternative than mutexes or creating getter/setter functions for a write-once scenario like this. I was looking for an alternative that would be simple and does not require us to remember locking the mutex.

test/integration/dbservice/dbservice.go
```diff
@@ -37,7 +39,10 @@ func (dbs *dbService) Stop(ctx context.Context) {
...
+	defer dbs.driverMu.Unlock()
+
 	if dbs.driver != nil {
 		return dbs.driver
 	}
```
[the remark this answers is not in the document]
keremgocen: Copilot is actually incorrect here; the mutex does prevent multiple initialisations. But it's less stylish and less idiomatic apparently:)

claude says:
```                                                                                                                         
  With mutex (current code):                                                                                             
  Goroutine 1: Lock() → acquired                                                                                         
  Goroutine 2: Lock() → blocked, waiting...                                                                              
  Goroutine 1: Check driver == nil? Yes → Create driver → Set driver → Unlock                                            
  Goroutine 2: Lock() → acquired (now gets lock)                                                                         
  Goroutine 2: Check driver == nil? NO (already set!) → Return existing driver → Unlock                                  
                                                                                                                         
  The mutex ensures only ONE goroutine can be inside the Lock()...Unlock() section at a time, so the check and           
  initialization are atomic.                                                                                             
                                                                                                                         
  HOWEVER, Copilot has a good point about style: sync.Once is more idiomatic for one-time initialization: 
```

we can update this to use sync.Once instead, I'll leave it for review

PR #156, opened 2026-01-15
Enable gds for http

MacondoExpress:
This PR adds support for GDS when the server starts in HTTP mode.
Prior to this PR, when HTTP mode was introduced, the checks included in verifyRequirements were skipped, due to the fact that DBMS credentials required to perform them were not available as passed by MCP clients as basic authorization headers.
This PR solves this issue by introducing the "BeforeInitialize" hook, which intercepts the MCP client request containing the basic authentication, allowing the server to perform these requirements checks once the credentials are available. 
Implementation details:
- Using a flag "httpConnectionVerified," the server performs these checks only once during the server lifecycle.
If the flag "httpConnectionVerified" is already set to true, the hooks avoid being performed again, avoiding unnecessary roundtrips between the server and the DBMS.
- If GDS is found during the "verifyRequirements," then it is added to the list of tools.
- "StartupEvent" is now sent, with all the details as for the STDIO version.
- Differently from STDIO mode, errors on "verifyRequirements" do not result in the server stopping. This is because "verifyRequirements" could fail due to invalid credentials.

### DIFF
--- a/internal/server/server_http_lifecycle_test.go
+++ b/internal/server/server_http_lifecycle_test.go
@@ -44,7 +44,7 @@ func TestNeo4jMCPServerHTTPMode(t *testing.T) {
 		Database:      "neo4j",
 		TransportMode: config.TransportModeHTTP,
 		HTTPHost:      "127.0.0.1",
-		HTTPPort:      "8080",
+		HTTPPort:      "3000",
 	}
 	uri := fmt.Sprintf("http://%s:%s/mcp", cfg.HTTPHost, cfg.HTTPPort)
 


internal/server/server.go
```diff
@@ -32,60 +33,89 @@ const (
...
+			slog.Info("Verify server requirements...")
+			if err := s.verifyRequirements(ctx); err != nil {
+				slog.Error("Error during the requirements verification:", "error", err)
+				return
+			}
+			if s.gdsInstalled {
+				s.addGDSTools()
+			}
+			s.emitStartupEvent(ctx)
+			s.httpConnectionVerified = true
```
keremgocen: we should also use `sync/atomic` here as you'll hit the same race condition issue I had earlier. (you can see it if you run the tests with the race flag `go test ./... -race`)

internal/server/server.go
```diff
@@ -32,60 +33,89 @@ const (
...
+		// in case of http mode, the initialization process is delayed until the credentials are available.
+		// when the first client is performing the initialize request then the server perform
+		s.hooks.AddBeforeInitialize(func(ctx context.Context, _ any, _ *mcp.InitializeRequest) {
+			// if the connection is already verified does not perform the check anymore.
+			if s.httpConnectionVerified {
+				return
+			}
+
+			slog.Info("Verify server requirements...")
+			if err := s.verifyRequirements(ctx); err != nil {
```
keremgocen, line 104: not necessarily in this PR, but I'm now thinking if we should also gather the connection info inside `verifyRequirements` while we are here. Previously, this was skipped for the HTTP mode.

We are sending another db query to gather connection info if the telemetry is enabled, so maybe it makes sense to get that info here 🤔

internal/server/server.go
```diff
@@ -32,60 +33,89 @@ const (
...
-
 	switch s.config.TransportMode {
 	case config.TransportModeHTTP:
+		slog.Info("Registering server tools")
+		if err := s.registerTools(); err != nil {
+			return err
+		}
+		// in case of http mode, the initialization process is delayed until the credentials are available.
+		// when the first client is performing the initialize request then the server perform
+		s.hooks.AddBeforeInitialize(func(ctx context.Context, _ any, _ *mcp.InitializeRequest) {
```
keremgocen, line 89: I think this looks pretty cool, we should use hooks more as they provide a lot of convenience.

internal/server/server.go
```diff
@@ -32,60 +33,89 @@ const (
...
+			err := s.verifyRequirements(context.Background())
+			if err != nil {
+				return err
+			}
+
+			// Register tools
+			if err := s.registerTools(); err != nil {
+				return fmt.Errorf("failed to register tools: %w", err)
+			}
+			s.emitStartupEvent(context.Background())
```
keremgocen, line 128: I was thinking we can move the event sending outside the switch since they are shared, any reason to use a different context here?

internal/server/server_http_lifecycle_test.go
```diff
@@ -0,0 +1,498 @@
+// Package server_test contains tests for the HTTP mode lifecycle of the Neo4j MCP server.
+// This file specifically tests the differences between HTTP and STDIO transport modes,
+// with focus on the delayed initialization pattern used in HTTP mode.
+//
+// Key differences tested:
+// - HTTP mode: Verification and tool registration are delayed until the first client initializes
+// - STDIO mode: Verification and tool registration happen immediately during Start()
+//
+// The HTTP mode uses hooks to defer the initialization process because:
+// - Database credentials are provided per-request via Basic Auth headers
```
keremgocen: and bearer now, we now have Bearer token as well in main

internal/server/server_http_lifecycle_test.go
```diff
@@ -0,0 +1,498 @@
...
+
+	})
+
+	t.Run("Server handles database connectivity errors gracefully", func(t *testing.T) {
+		mockDB := db.NewMockService(ctrl)
+		// in HTTP the serve should keep running even if the connectivity check fails.
+		// This is because the client can be misconfigured with invalid credentials
+		// and it should no affect the experience to other clients/users with correct information.
+
+		mockDB.EXPECT().VerifyConnectivity(gomock.Any()).Times(1).Return(fmt.Errorf("connection error"))
```
keremgocen, line 115: I'm a bit on the fence about this, as we discussed earlier. I think if the verification fails for some reason, the client should be informed with an error. Not shutdown the server but we should return an error to the initialize request I think so the client is aware, instead of sliently failing. That would be a better user experience imo, what do you think?

internal/server/server_http_lifecycle_test.go
```diff
@@ -0,0 +1,498 @@
...
+		errChan := make(chan error, 1)
+		go func() {
+			err := s.Start()
+			if err != nil {
+				errChan <- err
+			}
+		}()
+
+		// Give the server a moment to start
+		time.Sleep(100 * time.Millisecond)
```
keremgocen: we have many tests using this sleep, but we already have a `httpServerReady` channel in the server we could use for this purpose. Although you'll have to export it, I think. This feels a bit brittle currently.

internal/server/server_http_lifecycle_test.go
```diff
@@ -0,0 +1,498 @@
...
+			t.Fatalf("error while initialize request: %v", err)
+		}
+		// Expected tools that should be registered
+		// update this number when a tool is added or removed.
+		// All tools: get-schema, read-cypher, write-cypher, list-gds-procedures
+		expectedTotalToolsCount := 4
+
+		registeredTools := len(s.MCPServer.ListTools())
+
+		if expectedTotalToolsCount != registeredTools {
```
keremgocen: instead of checking the tool count, I think we can use the client to get the tool list and check specific tools. Which should help verify the registration as well?

something like this should work:
```
toolNames := make([]string, 0, len(s.MCPServer.ListTools()))                                                           
for _, tool := range s.MCPServer.ListTools() {                                                                         
    toolNames = append(toolNames, tool.Name)                                                                           
}                                                                                                                      
assert.Contains(t, toolNames, "get-schema")                                                                            
assert.Contains(t, toolNames, "list-gds-procedures") // when GDS installed
```

internal/server/server_http_lifecycle_test.go
```diff
@@ -0,0 +1,498 @@
...
+		// The hook will handle errors when actually triggered by a client request
+		s := server.NewNeo4jMCPServer("test-version", cfg, mockDB, analyticsService)
+
+		if s == nil {
+			t.Fatal("NewNeo4jMCPServer() returned nil")
+		}
+
+		// Start HTTP server in goroutine
+		errChan := make(chan error, 1)
+		go func() {
```
keremgocen: server setup seems to be shared across all tests
- start
- wait 100 ms
- stop
- check err channel

should we move that logic into a helper function, I think something like `setupHTTPServer` would help clean things up a bit, what do you think?

internal/server/server_http_lifecycle_test.go
```diff
@@ -0,0 +1,498 @@
...
+		case err := <-errChan:
+			if err != nil {
+				t.Errorf("Start() should not fail due to hook errors = %v", err)
+			}
+		default:
+			// No error is expected
+		}
+	})
+
+	t.Run("Server should no perform duplicate verification calls", func(t *testing.T) {
```
keremgocen:
```suggestion
	t.Run("Server should not perform duplicate verification calls", func(t *testing.T) {
```

internal/server/server.go
MacondoExpress: Good point!

internal/server/server.go
MacondoExpress: `ExecuteReadQuery` uses credentials from the basic auth so it requires the HTTP request context, while in STDIO mode the driver is already initialised with auth

internal/server/server_http_lifecycle_test.go
MacondoExpress: a good catch, I have to sync the branch then!

internal/server/server_http_lifecycle_test.go
MacondoExpress: I agree, not as part of this PR, as it's already the current behaviour and it's outside from the scope which is to enable GDS for http users. Also it will requires more design, I'm unsure if we can use the initialise request to communicate these information to the MCP client

internal/server/server_http_lifecycle_test.go
MacondoExpress: Better!

internal/server/server_http_lifecycle_test.go
MacondoExpress: Done!

internal/server/server.go
MacondoExpress: I'll wait the other PR to be merged, so that I can harmonise the solution with the other changes

PR #164, opened 2026-01-20
move minor patch notes as separate entries

MacondoExpress:
as title


.changes/unreleased/Minor-20260120-133310.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Minor
+body: Analytics events now also emitted when the server is executed in HTTP mode; previously, only STDIO mode supported Telemetry. (needs to be enabled explicit
```
mjfwebb:
```suggestion
body: Telemetry events are now also emitted when the server is executed in HTTP mode. Previously only STDIO mode supported telemetry.
```

Issue #169, opened 2026-01-29
Consistency among flags and environment variables for transport mode


luffy1727:
Problem: Environment variable `NEO4J_MCP_TRANSPORT` counterpart for cli flag is named `--neo4j-transport-mode` which introduces inconsistency in naming.

Solution: Rename the env variable to `NEO4J_TRANSPORT_MODE`

PR #170, opened 2026-01-29
adding support for env variable NEO4J_TRANSPORT_MODE

luffy1727:
Issue [169](https://github.com/neo4j/mcp/issues/169)


internal/cli/args.go
```diff
@@ -26,7 +26,7 @@ Options:
   --neo4j-read-only <BOOLEAN>         Enable read-only mode: true or false (overrides environment variable NEO4J_READ_ONLY)
   --neo4j-telemetry <BOOLEAN>         Enable telemetry: true or false (overrides environment variable NEO4J_TELEMETRY)
   --neo4j-schema-sample-size <INT>    Number of nodes to sample for schema inference (overrides environment variable NEO4J_SCHEMA_SAMPLE_SIZE)
-  --neo4j-transport-mode <MODE>       MCP Transport mode (e.g., 'stdio', 'http') (overrides environment variable NEO4J_MCP_TRANSPORT)
+  --neo4j-transport-mode <MODE>       MCP Transport mode (e.g., 'stdio', 'http') (overrides environment variable NEO4J_TRANSPORT_MODE)
```
keremgocen: we don't want to change the existing NEO4J_MCP_TRANSPORT but to add a new NEO4J_TRANSPORT_MODE to provide consistency with the existing flag `neo4j-transport-mode` was my understanding from earlier discussions. 
And at a later (major) release we want to deprecate the old `NEO4J_MCP_TRANSPORT` so we should clearly document it as deprecated, both in code and in the docs as well. (cc: @MacondoExpress @mjfwebb to confirm)

MacondoExpress: Correct,
@luffy1727 this line changed looks good.

internal/cli/args.go
```diff
@@ -44,7 +44,7 @@ Optional Environment Variables:
   NEO4J_TELEMETRY Enable/disable telemetry (default: true)
   NEO4J_READ_ONLY Enable read-only mode (default: false)
   NEO4J_SCHEMA_SAMPLE_SIZE Number of nodes to sample for schema inference (default: 100)
-  NEO4J_MCP_TRANSPORT MCP Transport mode (e.g., 'stdio', 'http') (default: stdio)
+  NEO4J_TRANSPORT_MODE MCP Transport mode (e.g., 'stdio', 'http') (default: stdio)
```
MacondoExpress, line 47: You can keep both here, describing NEO4J_MCP_TRANSPORT as deprecated in favour of `NEO4J_TRANSPORT_MODE`

internal/cli/args.go
luffy1727: I don't see the point of supporting 2 different variables in the code. Support for `NEO4J_MCP_TRANSPORT` hasn't been dropped. Docs are updated to indicate that `NEO4J_MCP_TRANSPORT` variable is deprecated.

MacondoExpress: > I don't see the point of supporting 2 different variables in the code. Support for `NEO4J_MCP_TRANSPORT` hasn't been dropped. Docs are updated to indicate that `NEO4J_MCP_TRANSPORT` variable is deprecated.

We want user to be able to smooth migrate between major versions, and supporting two environment variables while deprecating one is the SemVer way to do it in this case,
for reference: https://semver.org/#how-should-i-handle-deprecating-functionality

luffy1727: Sorry for making it confusing. The backward compatibility is still there and it supports both variables. Maybe I misunderstood but I don't understand what @keremgocen referencing here.

MacondoExpress: I'm not sure as well, there is nothing particular wrong in this line IMO.

keremgocen: Sorry, this is correct as it's only changing the comment on the existing flag.

I wrote my initial comment because I thought this PR was replacing the old env variable with the new one, and I tried to say "we want to keep both env variables instead of replacing the old one". 

So elsewhere in the PR, we want both `NEO4J_MCP_TRANSPORT` and the new `NEO4J_TRANSPORT_MODE` to co-exist until the next major release. And we want to document this behaviour in comments.

test/integration/deprecation_test.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/neo4j/mcp/test/dbservice"
+)
+
+var testDbs = dbservice.NewDBService()
+
+func TestIfWeStillSupportVersionOne(t *testing.T) {
```
keremgocen: how will this be used? It does not look like it will ever fail unless we change the test code in `startOnce`

.changes/unreleased/Minor-20260203-100801.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Minor
+body: deprecating NEO4J_MCP_TRANSPORT env variable
```
MacondoExpress:
```suggestion
The NEO4J_MCP_TRANSPORT environment variable is now deprecated in favour of NEO4J_TRANSPORT_MODE.
```

test/containerrunner/container_runner.go
```diff
@@ -51,18 +51,27 @@ func GetDriverConf() *config.Config {
 }
 
 // startOnce start the testcontainer imaged
-func startOnce(ctx context.Context) {
+func startOnce(ctx context.Context, version string) {
```
MacondoExpress: The following config are only required for starting the Neo4j database and driver, you can unit test the behaviour directly on the unit tests in config_test.go. 
I would revert this test changes in favour of the unit test in config_test.go

MacondoExpress (changes requested):
Left some comments, 
I would change the test strategy used in favour of a simple unit test here.

test/containerrunner/container_runner.go
luffy1727: removed

Issue #173, opened 2026-01-30
AWS AgentCore health check fails: MCP ping requires auth and rejects /mcp/


Andy2003:
**Description:**

`ping` cannot be used as an unauthenticated health check against the Neo4j MCP HTTP transport. This breaks AWS AgentCore runtime health checks (and also fails locally), because:

1) The server does not accept a trailing slash on the MCP path (`/mcp/`), returning `404 Not Found` (likely due to `pathValidationMiddleware`).
2) The JSON-RPC `ping` request is blocked by auth requirements when sent without an `Authorization` header, returning `401 Unauthorized: Basic or Bearer authentication required`.

According to the MCP spec, `ping` is a utility method intended to be callable for connectivity checks:
https://modelcontextprotocol.io/specification/2025-03-26/basic/utilities/ping

**How to Reproduce:**

This can be reproduced with a **local Docker** MCP server.

1) Start the container:

```bash
docker run --rm --name neo4-mcp \
  -e NEO4J_URI=neo4j+s://demo.neo4jlabs.com:7687 \
  -e NEO4J_DATABASE=companies \
  -e NEO4J_MCP_TRANSPORT=http \
  -e NEO4J_MCP_HTTP_HOST=0.0.0.0 \
  -e NEO4J_MCP_HTTP_PORT=8000 \
  -p 8000:8000 \
  mcp/neo4j
````

2. Send a `ping` request to `/mcp` (no auth header):

*Request*:

```http
POST http://localhost:8000/mcp
Content-Type: application/json
Accept: application/json

{"jsonrpc":"2.0","method":"ping","params":null,"id":4}
```

*Response*:

```http
HTTP/1.1 401 Unauthorized
Content-Type: text/plain; charset=utf-8
Www-Authenticate: Basic realm="Neo4j MCP Server"
Www-Authenticate: Bearer realm="Neo4j MCP Server"
X-Content-Type-Options: nosniff
Content-Length: 54

Unauthorized: Basic or Bearer authentication required
```

3. Send the same request to `/mcp/` (trailing slash):

*Request*:

```http
POST http://localhost:8000/mcp/
Content-Type: application/json
Accept: application/json

{"jsonrpc":"2.0","method":"ping","params":null,"id":4}
```

*Response*:

```http
HTTP/1.1 404 Not Found
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
Content-Length: 53

Not Found: This server only handles requests to /mcp
```

[...]

PR #175, opened 2026-01-30
Add support for custom auth header configuration

Andy2003:
This pull request adds support for configuring custom authorization headers. This change allows greater flexibility in authentication mechanisms by enabling users to define their own headers.

resolves #174

PR #179, opened 2026-02-03
Allow unauthenticated "ping" requests and support "/mcp/" path in middleware

Andy2003:
This PR adds support for unauthenticated "ping" health check requests in the authentication middleware.
When no credentials are provided, the middleware now reads the request body to check if it's a JSON-RPC ping method call, and allows it through without authentication.

This enables external systems like load balancers or monitoring tools to perform liveness checks without requiring authentication credentials.

resolves half of #173

### DIFF
--- a/internal/cli/args.go
+++ b/internal/cli/args.go
@@ -71,7 +71,7 @@ For more information, visit: https://github.com/neo4j/mcp
 type Args struct {
 	URI                          string
 	Username                     string
-	Password                     string
+	Password                     string // #nosec G117 -- Password is only used during startup to create auth token, not logged or exposed
 	Database                     string
 	ReadOnly                     string
 	Telemetry                    string
@@ -104,7 +104,7 @@ var argsSlice = []string{
 	"--neo4j-http-tls-enabled",
 	"--neo4j-http-tls-cert-file",
 	"--neo4j-http-tls-key-file",
-	"--neo4j-auth-header-name",
+	"--neo4j-http-auth-header-name",
 	"--neo4j-http-allow-unauthenticated-ping",
 }
 
--- a/internal/server/middleware.go
+++ b/internal/server/middleware.go
@@ -65,8 +65,7 @@ func authMiddleware(headerName string, allowUnauthenticatedPing bool) func(http.
 			authHeader := r.Header.Get("Authorization")
 
 			// Try the bearer token first
-			if strings.HasPrefix(authHeader, "Bearer ") {
-				token := strings.TrimPrefix(authHeader, "Bearer ")
+			if token, found := strings.CutPrefix(authHeader, "Bearer "); found {
 				token = strings.TrimSpace(token)
 
 				if token == "" {
@@ -190,7 +189,7 @@ func loggingMiddleware() func(http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 
-			slog.Debug("HTTP Request",
+			slog.Debug("HTTP Request", // #nosec G706 -- logging HTTP request metadata, no user input in format string
 				"method", r.Method,
 				"url", r.URL.Path,
 				"remote_addr", r.RemoteAddr,
--- a/.changes/unreleased/Patch-20260218-234019.yaml
+++ b/.changes/unreleased/Patch-20260218-234019.yaml
@@ -1,0 +1,3 @@
+kind: Patch
+body: Fixed the type in argSlice preventing neo4j-http-auth-header-name from being parsed as expected.
+time: 2026-02-18T23:40:19.247917Z
--- a/internal/cli/args_test.go
+++ b/internal/cli/args_test.go
@@ -266,6 +266,19 @@ func TestHandleArgs(t *testing.T) {
 			expectedStderr:   "--neo4j-http-allowed-origins requires a value",
 		},
 		{
+			name:             "http auth header name flag with valid value",
+			args:             []string{testProgramName, "--neo4j-http-auth-header-name", "X-Custom-Auth"},
+			version:          testVersion,
+			expectedExitCode: -1, // Should not exit, flag is allowed
+		},
+		{
+			name:             "http auth header name flag missing value",
+			args:             []string{testProgramName, "--neo4j-http-auth-header-name"},
+			version:          testVersion,
+			expectedExitCode: 1,
+			expectedStderr:   "--neo4j-http-auth-header-name requires a value",
+		},
+		{
 			name:             "double dash separator stops flag processing",
 			args:             []string{testProgramName, "--", "--unknown-flag"},
 			version:          testVersion,
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -29,7 +29,7 @@ var ValidTransportModes = []TransportMode{TransportModeStdio, TransportModeHTTP}
 type Config struct {
 	URI                      string
 	Username                 string
-	Password                 string
+	Password                 string // #nosec G117
 	Database                 string
 	ReadOnly                 bool // If true, disables write tools
 	Telemetry                bool // If false, disables telemetry
@@ -104,7 +104,7 @@ func (c *Config) Validate() error {
 type CLIOverrides struct {
 	URI                      string
 	Username                 string
-	Password                 string
+	Password                 string // #nosec G117
 	Database                 string
 	ReadOnly                 string
 	Telemetry                string
--- a/internal/logger/logger_test.go
+++ b/internal/logger/logger_test.go
@@ -162,7 +162,7 @@ func TestLogLevelChange(t *testing.T) {
 
 func TestRedactionLogic(t *testing.T) {
 	t.Run("sensitive keys are redacted", func(t *testing.T) {
-		sensitiveFields := map[string]string{
+		sensitiveFields := map[string]string{ // #nosec G101 -- test data for redaction logic verification
 			"password":   "my-secret-password",
 			"token":      "bearer-token-123",
 			"api_key":    "sk-1234567890",


Andy2003:
@MacondoExpress How do we proceed with this PR?

internal/server/middleware.go
```diff
@@ -77,6 +80,35 @@ func authMiddleware(headerName string) func(http.Handler) http.Handler {
 			// Fall back to basic auth
 			user, pass, ok := r.BasicAuth()
 			if !ok {
+				// No credentials provided - check for unauthenticated ping health check
+				// Allow a JSON-RPC ping to pass through unauthenticated for liveness probes
+				// but ONLY when the request is a POST to /mcp or /mcp/ and the body contains
+				// a JSON object with "method":"ping". This keeps other RPC methods protected.
+
+				const maxBodyBytes = 4 * 1024 * 1024 // 4MB guard
```
MacondoExpress: I would reduce this guard to something more sensible as we would expect to contains only a JSON-RPC with method "ping"

MacondoExpress (changes requested):
Can you move this feature behind feature flag?

.changes/unreleased/Patch-20260205-153619.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Patch
+body: 'Allow unauthenticated "ping" requests'
```
MacondoExpress: I would provide an hint of the env variable and/or arg that enable it in the changeset

MacondoExpress (commented):
@Liam-Doodson @keremgocen @mjfwebb  
This is now looking good to me,
Please have a look as well

.changes/unreleased/Patch-20260205-153619.yaml
```diff
@@ -0,0 +1,5 @@
+kind: Patch
+body: |
+  Allow unauthenticated "ping" requests. Enable with the `NEO4J_HTTP_ALLOW_UNAUTHENTICATED_PING` environment
+  variable or the `--neo4j-http-allow-unauthenticated-ping` CLI flag.
```
keremgocen, line 4: going forward, I think we want these env vars to be MCP-scoped to avoid any inconsistency and breaking changes. 
I'm leaning towards putting `mcp` somewhere in these.

So something like  `NEO4J_MCP_HTTP_ALLOW_UNAUTHENTICATED_PING` might work better. We have a few other rogue variables like this (without `mcp` in the name), but hopefully we'll fix those in the next major release.

docs/CLIENT_SETUP.md
```diff
@@ -101,6 +101,7 @@ neo4j-mcp
 export NEO4J_MCP_HTTP_HOST="127.0.0.1"      # Default: 127.0.0.1
 export NEO4J_MCP_HTTP_PORT="80"             # Default: 80
 export NEO4J_MCP_HTTP_ALLOWED_ORIGINS="*"   # Default: empty (no CORS)
+export NEO4J_HTTP_ALLOW_UNAUTHENTICATED_PING="false" # Allow unauthenticated ping probes (default: false)
```
keremgocen, line 104:
```suggestion
export NEO4J_MCP_HTTP_ALLOW_UNAUTHENTICATED_PING="false" # Allow unauthenticated ping probes (default: false)
```

internal/cli/args.go
```diff
@@ -54,6 +55,7 @@ Optional Environment Variables:
   NEO4J_MCP_HTTP_TLS_CERT_FILE Path to TLS certificate file (required when TLS is enabled)
   NEO4J_MCP_HTTP_TLS_KEY_FILE Path to TLS private key file (required when TLS is enabled)
   NEO4J_HTTP_AUTH_HEADER_NAME Name of the HTTP header to read auth credentials from (default: Authorization)
+  NEO4J_HTTP_ALLOW_UNAUTHENTICATED_PING Allow unauthenticated ping health checks (default: false)
```
keremgocen, line 58:
```suggestion
  NEO4J_MCP_HTTP_ALLOW_UNAUTHENTICATED_PING Allow unauthenticated ping health checks (default: false)
```

internal/server/middleware.go
```diff
@@ -178,3 +186,28 @@ func loggingMiddleware() func(http.Handler) http.Handler {
...
+
+func isUnauthenticatedPingRequest(r *http.Request) bool {
+	if r.Method != http.MethodPost {
+		return false
+	}
+	if r.ContentLength >= 0 && r.ContentLength > maxUnauthenticatedBodyBytes {
+		return false
+	}
+
+	buf, err := io.ReadAll(r.Body)
```
keremgocen, line 216: please correct me if I'm wrong but using `io.ReadAll` here will ignore our intentions of limiting the content size we want to read, if the client does not send the `Content-Lenght` header, so this becomes an unbounded reader making things even worse.

So we can either wrap the current request body with an `io.LimitReader` or use something like [http.MaxBytesReader](https://pkg.go.dev/net/http#MaxBytesReader) to ensure we restrict the limit no matter the transfer encoding.

internal/server/middleware_test.go
```diff
@@ -612,3 +613,73 @@ func TestPathValidationMiddleware_TrailingSlashAllowed(t *testing.T) {
...
+		t.Errorf("Expected status 401, got %d", rec.Code)
+	}
+
+	// Should have WWW-Authenticate header
+	if rec.Header().Get("WWW-Authenticate") == "" {
+		t.Error("Expected WWW-Authenticate header to be set")
+	}
+}
+
+func TestAuthMiddleware_PassesBasicAuthToContext(t *testing.T) {
```
keremgocen: this test looks identical to the `TestAuthMiddleware_WithValidBasicCredentials` so we can drop this one.

internal/server/middleware.go
```diff
@@ -44,7 +48,7 @@ func (s *Neo4jMCPServer) chainMiddleware(allowedOrigins []string, next http.Hand
 // Credentials are extracted and stored in the request context for tools to create
 // per-request Neo4j driver connections, enabling multi-tenant scenarios.
 // Returns 401 Unauthorized if credentials are missing or malformed.
-func authMiddleware(headerName string) func(http.Handler) http.Handler {
+func authMiddleware(headerName string, allowUnauthenticatedPing bool) func(http.Handler) http.Handler {
```
keremgocen, line 54: I'm not sure about changing this function's signature, which impacts all other usages. 

Maybe not in this PR but maybe we could make authMiddleware a method on *Neo4jMCPServer similar to the chainMiddleware to get access to the s.config. I think then we would not need to expand this any further with each new parameter. 🤔 But please ignore this for now.

internal/server/middleware.go
```diff
@@ -10,7 +13,8 @@ import (
 )
 
 const (
-	corsMaxAgeSeconds = "86400" // 24 hours
+	corsMaxAgeSeconds           = "86400" // 24 hours
+	maxUnauthenticatedBodyBytes = 4 * 1024
```
keremgocen, line 18: do we still want 4 MB? I'm thinking it should be much smaller?

keremgocen (changes requested):
Thanks for the PR @Andy2003 I think it's quite nice and makes us align with the spec even further.

I did point out a few bits to improve before merging. Do you think we can also add a unit test where the request is rejected when the client sends larger than allowed bytes? I think that would give us a bit more confidence since we are using `io.ReadAll` currently.

.changes/unreleased/Patch-20260205-153619.yaml
MacondoExpress: I explicitly asked @Andy2003 to move towards the `MCP unscoped one` we can change it holistically in the future however

Andy2003: according to @MacondoExpress and https://github.com/neo4j/mcp/issues/184 the name should be as is https://github.com/neo4j/mcp/pull/175#discussion_r2768334879

internal/server/middleware.go
Andy2003: it's 4 kb

keremgocen: my bad, you are right

.changes/unreleased/Patch-20260205-153619.yaml
keremgocen: we are trying to land on something consistent internally. Let me get back to you on this. You don't need to change anything right now. Sorry for the confusion.

README.md
```diff
@@ -78,6 +78,9 @@ In case where the standard HTTP header "Authorization" can't be used, it's possi
 
 See the [Client Setup Guide](docs/CLIENT_SETUP.md) for configuration instructions for both modes.
 
+**Health Check Ping**
+If you expose the HTTP endpoint to load balancers or uptime checks, enable unauthenticated ping probes with the `NEO4J_HTTP_ALLOW_UNAUTHENTICATED_PING` env var or `--neo4j-http-allow-unauthenticated-ping true`. When disabled (default), every request—including ping—requires auth.
+
```
MacondoExpress:
```suggestion
**Unauthenticated MCP Ping**
By default, the ping method is protected by standard authentication flows. However, because the MCP specification allows pings prior to initialization, some integrations (such as AWS AgentCore) rely on this optional method as an initial health check mechanism.

To improve integration compatibility with these platforms, you can exclude the ping method from authentication requirements:

Environment Variable: `NEO4J_HTTP_ALLOW_UNAUTHENTICATED_PING=true`

Command Line Flag: `--neo4j-http-allow-unauthenticated-ping true`

```

internal/server/middleware.go
```diff
@@ -81,9 +84,25 @@ func authMiddleware(headerName string, allowUnauthenticatedPing bool) func(http.
 			// Fall back to basic auth
 			user, pass, ok := r.BasicAuth()
 			if !ok {
-				if allowUnauthenticatedPing && isUnauthenticatedPingRequest(r) {
-					next.ServeHTTP(w, r)
-					return
+				if allowUnauthenticatedPing {
+					// Wrap the body; this will cause reads past the limit to fail. Only apply
+					// when the feature is enabled to avoid unintended side effects on other endpoints.
+					r.Body = http.MaxBytesReader(w, r.Body, maxUnauthenticatedBodyBytes)
```
keremgocen, line 89: 👍

internal/server/middleware.go
```diff
@@ -81,9 +84,25 @@ func authMiddleware(headerName string, allowUnauthenticatedPing bool) func(http.
...
+
+					ok, err := isUnauthenticatedPingRequest(r)
+					if err != nil {
+						// If the body was too large, return 413 Payload Too Large
+						if errors.Is(err, errRequestBodyTooLarge) {
+							http.Error(w, err.Error(), http.StatusRequestEntityTooLarge)
```
keremgocen, line 95: We don't see this code used often:) But this is much better, thanks!

keremgocen (approved):

PR #189, opened 2026-02-10
Remove unused MCPVersion and bump patch version

keremgocen:
Eliminate the unused MCPVersion field from the configuration and update the patch version to reflect these changes.

### DIFF
--- a/internal/server/server_test.go
+++ b/internal/server/server_test.go
@@ -296,8 +296,12 @@ func TestNewNeo4jMCPServerEvents(t *testing.T) {
 
 	t.Run("emits startup and OSInfoEvent and StartupEvent events on start", func(t *testing.T) {
 		analyticsService.EXPECT().IsEnabled().Times(1).Return(true)
-		analyticsService.EXPECT().NewStartupEvent(gomock.Any(), gomock.Any(), gomock.Any()).Times(1)
-		analyticsService.EXPECT().NewConnectionInitializedEvent(gomock.Any()).Times(1)
+		analyticsService.EXPECT().NewStartupEvent(config.TransportModeStdio, false, "test-version").Times(1)
+		analyticsService.EXPECT().NewConnectionInitializedEvent(analyticsReal.ConnectionEventInfo{
+			Neo4jVersion:  "5.18.0",
+			Edition:       "enterprise",
+			CypherVersion: []string{"5"},
+		}).Times(1)
 		analyticsService.EXPECT().EmitEvent(gomock.Any()).Times(2) // startup + connection events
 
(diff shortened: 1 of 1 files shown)


internal/server/server_test.go
```diff
@@ -315,3 +316,105 @@ func TestNewNeo4jMCPServerEvents(t *testing.T) {
...
+
+	t.Run("connection initialized event receives neo4j version, edition, and cypher versions", func(t *testing.T) {
+		analyticsService.EXPECT().IsEnabled().Times(1).Return(true)
+		analyticsService.EXPECT().NewStartupEvent(gomock.Any(), gomock.Any(), gomock.Any()).Times(1)
+		analyticsService.EXPECT().NewConnectionInitializedEvent(analyticsReal.ConnectionEventInfo{
+			Neo4jVersion:  "5.18.0",
+			Edition:       "enterprise",
+			CypherVersion: []string{"5", "25"},
+		}).Times(1)
+		analyticsService.EXPECT().EmitEvent(gomock.Any()).Times(2)
```
keremgocen: we seem to be sending this event twice (in other tests as well), and I'm not sure if that's right or was intentional?

internal/server/server_test.go
```diff
@@ -315,3 +316,105 @@ func TestNewNeo4jMCPServerEvents(t *testing.T) {
 		}
 	})
 }
+
+func TestStartupEventPassesCorrectFields(t *testing.T) {
```
MacondoExpress: You don't need this test, there is already a test above that is testing that the StartupEvent is created.
You need just to remove `gomock.Any()` in the argument in favour of the expected version

internal/server/server_test.go
MacondoExpress: Sorry I missed this comment, I think looking at the test itself I expect `.EmitEvent` will be invoked for both `NewStartupEvent` and `NewConnectionInitializedEvent` which seems right to me

keremgocen: ah you're right, I've read it as we are sending NewConnectionInitializedEvent twice for some reason, not the EmitEvent. My bad.

internal/server/server_test.go
keremgocen: looks like I abandoned this PR and did not receive any notifications either. I'll add a small card to track it and take a look at this.

internal/server/server_test.go
MacondoExpress: No probem!

internal/server/server_test.go
keremgocen: removed the new tests, and updated the existing values to be more specific. Hopefully it should work now.

PR #195, opened 2026-02-19
Add license checking workflow with helper comments

keremgocen:
Enhance the workflow by introducing a license header validation script and adding helpful comments for clarity. This ensures that all relevant files include the required Neo4j copyright header.

### DIFF
--- a/.changes/unreleased/Patch-20260224-112157.yaml
+++ b/.changes/unreleased/Patch-20260224-112157.yaml
@@ -1,3 +1,0 @@
-kind: Patch
-body: Add licence headers for Neo4j.
-time: 2026-02-24T11:21:57.661055Z


.github/workflows/licence-checker.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6
+        with:
+          go-version: "1.25.7"
+
+      - name: Install addlicense
+        run: go install github.com/google/addlicense@b289835c0741ba2fb92dd5f61af070e79c69a55e # v1.2
+
+      - name: Run addlicense
+        run: addlicense -f ./addlicense -check $(find . -name "*.go" -type f -print0 | xargs -0)
```
MacondoExpress, line 25: I think in this way it checks all the files, we can run the checks only for the files modified in the PR, this avoid newly created PR to raise noise on files unrelated to the PR that the user has open.

MacondoExpress:
@keremgocen sounds good, we can merge it also as it is as well. I would remove the patch notes as it does not requires to be documented

.github/workflows/licence-checker.yml
angrykoala: Not necessarily a problem, in the graphql lib we need to do this because addlicense cannot cope with too many files

.changes/unreleased/Patch-20260224-112157.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Patch
+body: Add licence headers for Neo4j.
```
angrykoala: Is this a public facing change that should be in the changelog?

keremgocen: Not really, I forgot to remove this.

.github/workflows/licence-checker.yml
MacondoExpress: true, I said as "it will be good to have" but I already approved it without it https://github.com/neo4j/mcp/pull/195#issuecomment-3951163628

MacondoExpress (approved):

PR #197, opened 2026-02-22
Update CLIENT_SETUP.md

shashivj:
Added instructions for Kiro IDE from Amazon, https://kiro.dev/

### DIFF
--- a/docs/CLIENT_SETUP.md
+++ b/docs/CLIENT_SETUP.md
@@ -315,6 +315,67 @@ For Neo4j Enterprise or Aura with SSO/OAuth configured:
 
 **Note:** Replace the bearer token with your actual OAuth/SSO token from your identity provider.
 
+## Amazon Kiro Configuration
+
+The Amazon Kiro (https://kiro.dev/) is an agentic IDE that helps you do your best work with features such as specs, steering, and hooks.
+
+The Kiro MCP configuration is managed through a JSON file. The interface supports two levels of configuration:
+
+- User-level: ~/.kiro/settings/mcp.json applies to all workspaces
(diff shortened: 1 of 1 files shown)


shashivj:
Please review and approve my PR.

shashivj:
@LackOfMorals , Thanks for your feedback , I will add instructions for stdio mode in next PR. Please approve this PR, then I can work on your changes too.

docs/CLIENT_SETUP.md
```diff
@@ -315,6 +315,67 @@ For Neo4j Enterprise or Aura with SSO/OAuth configured:
...
+The Kiro MCP configuration is managed through a JSON file. The interface supports two levels of configuration:
+
+- User-level: ~/.kiro/settings/mcp.json applies to all workspaces
+- Workspace-level: .kiro/settings/mcp.json is specific to the current workspace
+
+You can apply the configuration using :
+
+- Open the command palette (use Cmd + Shift + P on Mac or Ctrl + Shift + P on Windows/Linux), search for MCP and select Kiro: Open workspace MCP config (JSON) or Kiro: Open user MCP config (JSON), depending on whether or not you want to configure the settings for the workspace or user level.
+
+### HTTP Mode
```
keremgocen, line 331: I'm actually thinking if we need this setup at all, or the stdio one if it's identical with the other clients (vscode, claude)
This bit of information looks duplicated and unnecessarily making the file even larger (we could just say see the existing HTTP config). 

Is there any Kiro specific config requirement other than the JSON file mentioned above?

If I remember correct, the STDIO setup between Vscode and Claude was slightly different, are there any gotchas for Kiro in STDIO mode for example? That could be a more valuable information to users other than duplicating the HTTP mode setup imo.

PR #209, opened 2026-02-27
Fix for when user cannot access neo4j , home / default db is not set

LackOfMorals:
In server.go,

- removed the execution of a simple query 

in service.go  
- VerifyConnectivity now tries to run a simple harmless query to test connectivity and access

### DIFF
--- a/internal/database/service.go
+++ b/internal/database/service.go
@@ -70,23 +70,23 @@ func (s *Neo4jService) buildQueryOptions(ctx context.Context, baseOptions ...neo
 }
 
 // VerifyConnectivity checks the driver can establish a valid connection with a Neo4j instance;
+// This is done by running a harmless test query against whatever database has been specified ( if any )
 func (s *Neo4jService) VerifyConnectivity(ctx context.Context) error {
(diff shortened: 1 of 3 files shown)


internal/database/service.go
```diff
@@ -70,23 +70,23 @@ func (s *Neo4jService) buildQueryOptions(ctx context.Context, baseOptions ...neo
 }
 
 // VerifyConnectivity checks the driver can establish a valid connection with a Neo4j instance;
+// This is done by running a harmless test query against whatever database has been specified ( if any )
 func (s *Neo4jService) VerifyConnectivity(ctx context.Context) error {
-	// For HTTP mode, extract credentials from context
-	if s.transportMode == config.TransportModeHTTP {
```
keremgocen, line 75: I'm thinking if this will cause an issue for the HTTP mode, since the driver will be missing auth in here. But looks like we are removing all the driver calls, which might be a bit misleading if we have any problem with the driver connection 🤔

MacondoExpress:
Closing this as superseded by the https://github.com/neo4j/mcp/pull/225

PR #220, opened 2026-03-04
Improve changie workflow handling for releases

keremgocen:
The workflow was failing when it's triggered on deletion of files in (.changes/unreleased/**) after a release PR is merged, and there's nothing to release.

- Added a manual count for unreleased files in `.changes/unreleased/ `, skips the step if the count is 0.
- The batch and other steps are only run if the count > 0. This ensures we don't run them unless there's anything to release.
- No more `error-on-continue` approach, which might hide real problems. 
- Removed the confusing continue-on-error comment
- Tested locally with act for both scenarios (with and without changes)

### DIFF
--- a/.github/workflows/changie.yml
+++ b/.github/workflows/changie.yml
@@ -7,32 +7,30 @@ on:
       - ".changes/unreleased/**"
 
 jobs:
+  check-unreleased:
+    runs-on: ubuntu-latest
+    outputs:
+      count: ${{ steps.check.outputs.count }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+      - id: check
+        run: echo "count=$(find .changes/unreleased -name "*.yaml" | wc -l)" >> $GITHUB_OUTPUT
+
   generate-pr:
+    needs: check-unreleased
+    if: needs.check-unreleased.outputs.count > 0
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
-
-      - name: Check for unbatched changes
-        id: check_changes
-        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/changie.yml
```diff
@@ -13,26 +13,41 @@ jobs:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
+      - name: Check for unbatched changes
```
MacondoExpress: My recommendation is to move this in an above job and then modify the generate-pr:
```
  generate-pr:
    needs: check-unreleased
    if: needs.check-unreleased.outputs.count != '0'
```
In this way the generate-pr job is skipped entirely and is cleaner to read, at least IMO
But, I would defer to @angrykoala as my main preference is that eventually `neo4j/cli` is aligned.

keremgocen: Not sure what you mean? This is already at the top of other steps, and it's run as the first step. 

Everything else is skipped if the batch fails, which is not run at all if the count from `check_changes` is 0.

mjfwebb: I think @MacondoExpress is basically saying for it to be an earlier exit if we do a single check earlier before the job starts.

Pseudocode:

```
do_thing = check_if_do_thing

if (do_thing) {
  This Job
}
```

instead of 

```
do_thing = check_if_do_thing

This job
  if (do_thing) {
     do_step()
  }
  if (do_thing) {
     do_step()
  }
  if (do_thing) {
     do_step()
  }
```

keremgocen: I added another job as suggested. From resources point of view we now have 2 runners (2 jobs) and do an extra checkout if the count > 0, but I guess it improves the readability and is better overall.

MacondoExpress: Thank you, it looks good to me!

MacondoExpress (approved):

PR #253, opened 2026-04-09
Update README to synergize with docs

rsill-neo4j:
also see https://github.com/neo4j/docs-mcp/pull/12


keremgocen (dismissed):
While I appreciate cleaner/concise documentation, this feels a bit too much at the same time to me. Please allow us some time to discuss this and do a thorough review.

CONTRIBUTING.md
Liam-Doodson: I think we're better off removing this section as we're not likely to keep it up to date. This should also just be covered by the main docs

```suggestion
```

Liam-Doodson: I also don't think this needs to be duplicated from the main docs and isn't specific to contributing so I think we should remove this too

Liam-Doodson: Looks like forks are enabled and this PR is actually coming from a fork so I'd at least get rid of that bit 👀 

I also don't think we need to tell people how to clone the repo though and should stick to info that is specific to this repo so should remove this section

Liam-Doodson: This is pretty mostly documenting the MCP protocol which is a bit overkill imo. At the very least I'd cut it down to just one example for basic and bearer auth but again those are fairly common patterns that I don't think need documenting here

Liam-Doodson: This comment probably spans both PRs but imo it's overkill for us to document how to generate TLS certs. There are many different ways to do it and that guide mainly focuses on self signed certs which is not something we should do in our main docs imo

I wouldn't be against a couple of examples here for self signed certs but I also don't think it's necessary

Liam-Doodson: I'd be slightly careful as this looks like it's recommending we return internal errors straight to users which isn't good practice

```suggestion
            return mcp.NewToolResultError("Invalid arguments"), nil
```

Liam-Doodson: Same as above, I'd remove the `err.Error()` as that looks like we're recommending returning internal errors straight to users

Liam-Doodson (commented):
Just added a few suggestions which is mainly just more docs we can cut from this repo imo

README.md
```diff
@@ -1,293 +1,25 @@
...
-- The ability to execute queries.
-- The presence of the APOC plugin.
-
-**HTTP Mode - Verification Skipped**
-In HTTP mode, startup verification checks are skipped because credentials come from per-request Basic Auth headers. The server starts immediately without connecting to Neo4j at startup.
-
-**Optional Requirements**
-If an optional dependency is missing, the server will start in an adaptive mode. For instance, if the Graph Data Science (GDS) library is not detected in your Neo4j installation, the server will still launch but will automatically disable all GDS-related tools, such as `list-gds-procedures`. All other tools will remain available.
-
-## Installation (Binary)
```
Liam-Doodson, line 35: Would be nice to have some kind of quickstart guide in the readme still imo

Maybe just:
1. How to install using brew (with link to an install page in the docs for other methods)
2. How to configure the server (with a generic config and links to specific configs for different clients)

CONTRIBUTING.md
rsill-neo4j: removed this section as well

rsill-neo4j: i guess that holds for the entire "environment variables" section. removing that as well

rsill-neo4j: removing

rsill-neo4j: removing

README.md
rsill-neo4j: turned out really short, see if that is what you suggested

docs/BUILD_MCPB.md
Liam-Doodson, line 1: Not sure what's happened here but it looks like the contents is deleted but the files are actually still left behind

README.md
```diff
@@ -1,293 +1,67 @@
...
-See the [Client Setup Guide](docs/CLIENT_SETUP.md) for configuration examples.
+* Explore your graph schema - discover node labels, relationship types, and property keys so the AI can reason about your data model without prior knowledge of it.
+* Run Cypher queries - execute, read, and write queries against your database in response to natural language prompts.
+* Inspect and analyze data - retrieve nodes, relationships, and paths to answer questions, generate summaries, or feed data to other workflows.
 
-### CLI Flags
+## Quickstart with Homebrew
 
-You can override any environment variable using CLI flags:
+On a Mac, do:
```
Liam-Doodson: Technically homebrew isn't mac specific and isn't installed on every mac

```suggestion
To install using homebrew, do:
```

docs/TLS_SETUP.md
Liam-Doodson, line 1: I think there's still some value in this guide in the context of docs for contributors

I'd probably just cut the scope slightly to not minimise descriptions of config options and to make clear this guide is for contributors manually testing the server, not for real users/deployments of the server

docs/BUILD_MCPB.md
rsill-neo4j: yes, i only deleted the content, will delete the files as well

docs/TLS_SETUP.md
rsill-neo4j: restored the file and added a note -
for reducing the scope, what should go?

Liam-Doodson: Just had another read through and I think let's just leave this as it is for the most part. I'll put some specific comments

docs/TLS_SETUP.md
```diff
@@ -1,14 +1,19 @@
...
+**Note**: This guide is meant for contributors who are manually testing the server, not for users and deployments of the server.
 
-### Certificate Format
+
+## Important certificate requirements
+
+
+### Certificate format
 
 All certificates must be in **PEM format** (text-based format with `-----BEGIN CERTIFICATE-----` headers). The server does not support other formats like DER or PKCS12.
```
Liam-Doodson: The main docs should be the source of truth for this

```suggestion
```

docs/TLS_SETUP.md
```diff
@@ -1,14 +1,19 @@
-# TLS/HTTPS Setup for Neo4j MCP Server
+# TLS/HTTPS setup for Neo4j MCP server
 
 This guide covers TLS/HTTPS configuration for the Neo4j MCP server, including certificate generation, testing, and production deployment.
```
Liam-Doodson:
```suggestion
This guide covers TLS/HTTPS configuration for contributors who are manually testing the Neo4j MCP server during development. It uses self-signed certificates which are not suitable for production deployments.
```

docs/TLS_SETUP.md
```diff
@@ -1,14 +1,19 @@
...
 
-## Important Certificate Requirements
+**Note**: This guide is meant for contributors who are manually testing the server, not for users and deployments of the server.
```
Liam-Doodson: If you use my suggestion above to set the scope of this guide, this is covered

```suggestion
```

docs/TLS_SETUP.md
Liam-Doodson: If we cut the scope to just self signed certs for manual testing this isn't needed

```suggestion
```

docs/TLS_SETUP.md
```diff
@@ -26,15 +31,17 @@ Self-signed certificates are only suitable for:
 
 See the [Production Use](#production-use) section below for proper setup.
 
-**Note**: Automated tests generate certificates dynamically. For manual testing or production deployment, follow the steps below.
+**Note**: Automated tests generate certificates dynamically.
+For manual testing or production deployment, follow the steps below.
```
Liam-Doodson: This guide should only be for manual testing

```suggestion
```

docs/TLS_SETUP.md
Liam-Doodson: Not sure if the suggested change is working for some reason 🤔

docs/TLS_SETUP.md
```diff
@@ -89,13 +96,16 @@ export NEO4J_MCP_HTTP_TLS_KEY_FILE="key.pem"
 ./bin/neo4j-mcp
 ```
```
Liam-Doodson: To keep it simple let's just do the one example 

```suggestion
```

Liam-Doodson: Again, not sure the suggested change is working...

docs/TLS_SETUP.md
```diff
@@ -203,4 +214,4 @@ curl -u neo4j:password https://your-domain.com/mcp \
 # Other paths will return 404
 curl -u neo4j:password https://your-domain.com/
 # Returns: "Not Found: This server only handles requests to /mcp"
-```
```
Liam-Doodson, line 206: This guide shouldn't contain any mention of production use if it's for internal contributors

README.md
Liam-Doodson: Yeah looks really good! Might be nice to show a quick usage example - I'll see if we can get a short vid/gif to show something like claude using the server to run a query

docs/TLS_SETUP.md
rsill-neo4j: just tested, i think it's because of the three backticks

README.md
```diff
@@ -1,293 +1,67 @@
...
 ```bash
-chmod +x neo4j-mcp
-sudo mv neo4j-mcp /usr/local/bin/
-```
-
-Windows (PowerShell / cmd):
-
-```powershell
-move neo4j-mcp.exe C:\Windows\System32
+pip install neo4j-mcp-server
```
Liam-Doodson, line 21: Not sure the best way to document this but when you pip install the server you need to run it in a different way...

The config later then needs to become something like:

```
{
...
  "command": "python",
  "args": ["-m", "neo4j_mcp_server"]
...
}
```

docs/TLS_SETUP.md
```diff
@@ -1,40 +1,28 @@
...
-- Your organization's internal CA
-- Commercial certificate providers
-
-Self-signed certificates are only suitable for:
-
-- Local development on `localhost`
-- Testing environments with relaxed security checks
-- Development scenarios where you control the client configuration
-
 See the [Production Use](#production-use) section below for proper setup.
```
Liam-Doodson: Can we remove anything about production use from this guide

```suggestion
```

docs/TLS_SETUP.md
```diff
@@ -173,7 +152,8 @@ openssl s_client -connect 127.0.0.1:8443 </dev/null 2>/dev/null | grep "Cipher"
 - **Content-Type**: MCP requests need `Content-Type: application/json` header
 - **Port**: Default port is 443 when TLS is enabled, 80 when TLS is disabled (configurable via `--neo4j-http-port` or `NEO4J_MCP_HTTP_PORT`)
 
-## Production Use
+
+## Production use
```
Liam-Doodson: Can we remove the production use section from this guide

README.md
rsill-neo4j: added to the VSCode listing

docs/TLS_SETUP.md
rsill-neo4j: removed

docs/TLS_SETUP.md
rsill-neo4j: removed

Liam-Doodson (approved):

keremgocen:
I'll port these over to #270

PR #261, opened 2026-04-15
make docker release job manual

luffy1727:


MacondoExpress (approved):

.github/workflows/docker-release.yml
```diff
@@ -1,6 +1,7 @@
 name: Docker Release
 
 on:
+  workflow_dispatch: {}
```
keremgocen, line 4: I think we also need to be able to pick the branch here as input so we can push v2.x? or does it come for free by default?

keremgocen: yeah, ignore my comment, there's a dropdown to pick the branch already

PR #267, opened 2026-04-15
make docker release job manual

luffy1727:
Make docker release job manual so that we can release a deployment on v2 branch


.github/workflows/docker-release.yml
```diff
@@ -11,7 +12,7 @@ env:
 
 jobs:
   get-version:
-    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    if: ${{ github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success' }}
```
keremgocen: I'm thinking if we would want to get the version as input to the manual workflow instead? Because if we get the latest version from changie, and if it's already pushed to Docker before, we will end up overwriting that image with the branch we trigger this job from. What do you think?

keremgocen (approved):

PR #274, opened 2026-04-22
Improve HTTP integration tests for database name parsing

keremgocen:
- Adds database name validation (HTTP) integration tests and cleans up method restriction ones
- Tests full chain with real DB via path /db/neo4j/mcp
- Some refactor of integration tests; replaced global dbs with singleton, moved StartHTTPServer into helpers.go, deleted the long-running TestServerLifecycle (which is covered in e2e and unit tests already)
- We now call cfg.Validate() in HTTP integration test setup config

In addition to tests;
- buildQueryOptions is also simplified, no longer returns an error                                      
- Added empty dbName guard in the context getter


MacondoExpress:
Can you split this PR in some smaller one? it's hard to follow the indirect reason of the changes as they are not scoped.
For instance I would add:
-Simply buildQueryOptions
-Add database name guard
and so on.. It will reduce the cognitive complexity for the reviewers, and it will reduce the time-to-merge!

MacondoExpress:
Shall we close this one? there was never a follow-up and likely it's outdated. I think the changes here are net positive and valid, a bit hard to follow but just because they are aggregated in the same PR

PR #275, opened 2026-04-23
fix: read-cypher and write-cypher tools/list inputSchema was empty due to incompatible jsonschema tags

MacondoExpress:
## Problem

`read-cypher` and `write-cypher` were advertising an empty `inputSchema` in `tools/list` — no `query`, no `params`. MCP clients (e.g. Claude Desktop) followed the empty schema and sent calls without a `query` argument, causing every invocation to fail with:

```
Query parameter is required and cannot be empty
```

## Root cause

The `jsonschema` struct tags on `ReadCypherInput` / `WriteCypherInput` used an `invopop/jsonschema`-style form:

```go
jsonschema:"default=MATCH(n) RETURN n,description=The Cypher query to execute"
```

`mcp-go` v0.46+ uses `google/jsonschema-go` to build input schemas. That library rejects any tag starting with `WORD=` and returns an error. `mcp.WithInputSchema` silently swallows that error, leaving the tool with the default empty schema.

## Fix

Switch the tags to the plain description form that `google/jsonschema-go` expects:

```go
jsonschema:"The Cypher query to execute"
```

Both `query` (required, string) and `params` (optional, object) are now correctly advertised in `tools/list`.

## Regression test

A new e2e subtest verifies that every tool returned by `tools/list` matches its spec: description, annotations (`readOnlyHint`, `destructiveHint`, `idempotentHint`, `openWorldHint`), input property names and JSON Schema types, and the `required` array.

Apart from that a new write_cypher_test.go was introduced in the E2E test suite to demonstrate that the default tag was not required.


luffy1727 (dismissed):

test/e2e/write_cypher_test.go
```diff
@@ -0,0 +1,62 @@
...
+		mcpClient.Close()
+	})
+
+	_, err = mcpClient.Initialize(ctx, helpers.BuildInitializeRequest())
+	require.NoError(t, err, "failed to initialize MCP server")
+
+	t.Run("write-cypher succeeds without params argument", func(t *testing.T) {
+		t.Parallel()
+		tc := helpers.NewE2ETestContext(t, dbs.GetDriver())
+		label := tc.GetUniqueLabel("NoPrams")
```
luffy1727: noParams maybe?

luffy1727 (approved):

PR #278, opened 2026-04-27
Add per-request driver and URI header parsing

keremgocen:
Implements a per-request driver registry and URI header resolver for handling Neo4j connections in HTTP mode. 

Breaking change for HTTP mode: NEO4J_URI is no longer required. Each request must supply `X-Neo4j-MCP-URI` with a valid bolt URI (bolt://, bolt+s://, neo4j://, etc.). STDIO mode is unchanged.

  - neo4jDriverMiddleware: resolves URI from header, creates a per-request driver, and closes it
  after the request
  - URIResolver: validates X-Neo4j-MCP-URI header and scheme
  - DriverRegistry: creates a fresh driver per call; auth is applied at query time via ExecuteQueryWithAuthToken
  - getDriver(ctx) in service layer: routes to context driver (HTTP) or struct driver (STDIO)
  - Config: `NEO4J_URI` required for STDIO only; driver creation in main.go gated on STDIO
  - `X-Neo4j-MCP-URI` added to CORS allowed headers


internal/config/config.go
```diff
@@ -72,9 +67,12 @@ func (c *Config) Validate() error {
 		return fmt.Errorf("invalid transport mode '%s', must be one of %v", c.TransportMode, ValidTransportModes)
 	}
 
-	// For STDIO mode, require username, password, and database from environment.
-	// For HTTP mode, credentials and database come per-request (Auth headers and URL path).
+	// For STDIO mode, require URI, username, password, and database from environment.
+	// For HTTP mode, credentials and database come per-request (Auth headers and URL path);
```
MacondoExpress:
```suggestion
	// For HTTP mode, URI, credentials and database come per-request (Auth headers and URL path);
```

internal/config/config.go
MacondoExpress, line 89: we need to validate the c.URI is not set for HTTP mode

MacondoExpress (commented):
Left some comments!

MacondoExpress:
I'll do another pass, leaved some comments for now, looks good!

MacondoExpress:
We need to add the changie file for the surface changes

internal/mcpcontext/mcpcontext.go
```diff
@@ -56,3 +57,15 @@ func HasAuth(ctx context.Context) bool {
 	_, okBearer := GetBearerToken(ctx)
 	return okBasic || okBearer
 }
+
+// WithDriver stores a Neo4j driver in the context
+// The driver is stored as any to avoid importing the neo4j driver package
+func WithDriver(ctx context.Context, driver any) context.Context {
```
MacondoExpress, line 63: It would be better drop the any type and add the type assertion both here and GetDriver

internal/server/uri_resolver_test.go
```diff
@@ -0,0 +1,111 @@
...
+		{
+			name:      "neo4j+ssc scheme",
+			setHeader: true,
+			header:    "neo4j+ssc://localhost:7687",
+			wantURI:   "neo4j+ssc://localhost:7687",
+		},
+		{
+			name:      "uri with credentials is passed through",
+			setHeader: true,
+			header:    "neo4j+s://user:pass@aura.example.com:7687",
```
MacondoExpress, line 66: This is an interested case, that to be fair I just think about it reading this test. This has to be verified and handled in some shape and form as currently we expect this credentials from separate and distinct headers, we should be sure to not log or log in redacted form when the URI is printed.

internal/server/middleware.go
```diff
@@ -70,6 +74,34 @@ func loggingMiddleware() func(http.Handler) http.Handler {
 	}
 }
 
+// neo4jDriverMiddleware resolves the Neo4j bolt URI from the header
+func neo4jDriverMiddleware(resolver URIResolver, registry database.DriverRegistry) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
MacondoExpress, line 80: something interesting here is that we are initialising the driver per-request, but this includes also mcp-spec-related requests that have no requirements for a driver to exists, on top of my head:
- `ping`
- `initialise`
- `tool/list`

MacondoExpress (changes requested):

internal/mcpcontext/mcpcontext.go
keremgocen: I think I disagree with this comment as it was a deliberate decision to keep mcpcontext loosely coupled and avoid making a utility package dependent on the neo4j driver. It shifts the responsibility to the caller (it has to be type asserted) but I think that's a reasonable trade-off.
One alternative could be to move this into server package, closer to where it's used which would be better. And this also part of the reason why I don't like utility packages in Go and prefer to duplicate small functionality like this, and move it as close as possible to where it's consumed.

internal/server/uri_resolver_test.go
keremgocen: added a redaction function for URLs, good call!

internal/server/middleware.go
keremgocen: I think this is a valid concern, but not in the scope of this PR. We could card it up though and take a proper look how to handle those cases imo.

internal/mcpcontext/mcpcontext.go
MacondoExpress: That's a fair take, and it's good to have a different opinion on this. I consider it a minor detail anyway, so not blocking. We can resolve this comment!
From my point of view, the project is hardly coupled with the Neo4j driver, so I personally think that decoupling and abstracting here would give little, if not zero, benefit while moving the type assertion overhead to the N consumers. It is also mentioned directly in the comments.
I could have, however, missed other reasons outside the abstraction vs concrete design choice.

internal/server/middleware.go
MacondoExpress: that's a sensible approach indeed!

luffy1727 (dismissed):

.changes/unreleased/Minor-20260506-230902.yaml
```diff
@@ -0,0 +1,8 @@
+kind: Minor
```
MacondoExpress: This is a Major

MacondoExpress (changes requested):

.changes/unreleased/Minor-20260506-230902.yaml
MacondoExpress:
```suggestion
kind: Major
```

MacondoExpress (approved):

luffy1727 (approved):

PR #283, opened 2026-05-05
Fix tool input properties and add e2e test (copied from upstream)

keremgocen:
Cherry-picking 226094a from master to fix input properties for read-cypher and write-cypher tools to ensure they are properly advertised in the list/tools response.

(leaving 90374cf, and ebca9df out as they will be merged eventually and not needed for the actual fix)


keremgocen:
This is required to fix the error "Query parameter is required and cannot be empty", which is already fixed in main with 1.5.2.

keremgocen:
closing this, I'll merge main back to 2.x to get the latest updates including this one

PR #287, opened 2026-05-11
Documentation updates for 2.x with migration guide

keremgocen:
- README is ported from main based on recent doc updates
- CONTRIBUTING.md - stripped down significantly, removed no longer relevant sections
- TLS_SETUP.md - updated for v2
- CLIENT_SETUP.md and BUILD_MCPB.md - both removed


docs/BUILD_MCPB.md
```diff
@@ -1,36 +0,0 @@
-## Build MCPB Bundle (for Claude Desktop)
```
MacondoExpress, line 1: Does makes to remove this? it's a contributing guide for developer that want to build their own mcpb and also for us when we build and *release* the mcpb. The only reason why we don't release it is that we didn't had a Cross-(OS/Arch) solution, which is no longer the case.
If the deletion is coming from a previous comment of mine, I meant that it should be removed from the user documentation: https://neo4j.com/docs/mcp/current/mcpb/  not the developer/contributing guide. 
I would definitely find helpful reading this when I want to build the MCPB.

CONTRIBUTING.md
```diff
@@ -354,22 +219,26 @@ When adding new database operations:
...
 - Tests failing unexpectedly → regenerate mocks, verify env vars, rerun full test suite.
 - Dependency/build issues → `go mod tidy`.
 
-## Update MCPB Bundle (for Claude Desktop)
+## Update the MCPB bundle (for Claude Desktop)
 
 If your changes impact the end-user configuration (e.g., adding new environment variables or modifying tool definitions), you must update the `manifest.json` file. This ensures that integrations like Claude Desktop are aware of the new server configuration.
 
-For more information refer to the dedicated guide: [the MCPB build documentation](docs/BUILD_MCPB.md).
+For more information refer to the dedicated guide: [the MCPB build documentation](https://neo4j.com/docs/mcp/current/installation#mcpb).
```
MacondoExpress, line 232: Same as above the right place for the mcpb building is in the source repository and not in the user manual.

MacondoExpress (changes requested):
Look good to me just left some comments

docs/BUILD_MCPB.md
keremgocen: I think we can discuss bringing back the mcpb and where it actually belongs to, but it's not this PR that's removing that. 
https://github.com/neo4j/mcp/tree/main/docs pretty much everything is removed in main already so this PR is basically reflecting all those changes + v2 updates, so I'd like to keep it scoped to that.

CONTRIBUTING.md
keremgocen: explained the scope above; possibly a separate PR we want to bring it back, if that's the consensus.

docs/BUILD_MCPB.md
MacondoExpress: Ah sorry I missed that this was not part of this PR, then it's fine

MacondoExpress (approved):

PR #289, opened 2026-05-11
Clarify read-only limitations in documentation

keremgocen:
- Enhance documentation for the `read-cypher` tool and `GetQueryType` function to clearly outline read-only limitations and responsibilities regarding query classification. 
- Update README for read-cypher limitations.
- Tool descriptions left unchanged.


internal/tools/cypher/read_cypher_spec.go
```diff
@@ -14,7 +14,7 @@ type ReadCypherInput struct {
 
 func ReadCypherSpec() mcp.Tool {
 	return mcp.NewTool("read-cypher",
-		mcp.WithDescription("read-cypher can run only read-only Cypher statements. For write operations (CREATE, MERGE, DELETE, SET, etc...), schema/admin commands, or PROFILE queries, use write-cypher instead."),
+		mcp.WithDescription("executes read-only Cypher queries that do not modify database data. Validation uses EXPLAIN and Neo4j's query classification—only queries classified as read-only ('r') are executed. Limitation: custom procedures/functions incorrectly classified as read-only by Neo4j may bypass this check. For write operations, schema/admin commands, or PROFILE queries, use write-cypher instead."),
```
MacondoExpress: I'm unsure if this information is helpful for the LLM, I would be tentative to modify and consume tokens in the context window for something that will be not beneficial for the LLM/Server iteration.
I think this should be helpful if is it brought to the visibility of the user in the user documentation rather than to the LLM.

keremgocen: Yeah makes sense, maybe we should leave this one as it was. We can think about an LLM friendly description but it's probably as good as saying "don't call this with queries that modify database state".

MacondoExpress: I would say so, this initiative started for the limitation point of view and therefore giving visibility to the user about this limitation. But here the users has no visibility, I would close this in favour of a PR on https://github.com/neo4j/docs-mcp

keremgocen: We can leave the tool descriptions, but I think it's still worth documenting the function docs and updating the Readme for read-cypher as they are user/programmer facing. Or would you prefer to keep that information in docs-mcp only?

I reasoned that if I were a consumer of this repo, it would be nice to know and not assume everything we send to `read-cypher` is always read-only, as that may not be the case. It's unlikely I'll read everything in docs-mcp before I try to run my MCP server locally, so I guess it depends on how much visibility we want to give to this concern.

I'll create a separate docs-mcp PR to document the risk in there as well.

MacondoExpress: The other changes, and eventually other code comments that could guide the developer are fine by me, even appreciated, I would just remove/revert from `mcp.WithDescription` which is used by a non human interpreter

MacondoExpress: Therefore I would remove as well the Changie as there will be no surface changes

keremgocen: done!

MacondoExpress (approved):

PR #292, opened 2026-05-18
[multi-tenant] move verifyRequiments per-request, enable all tools by default.

MacondoExpress:
Moves Neo4j connectivity and plugin checks from server startup to the MCP Initialize hook, so HTTP multi-tenant mode can validate each environments independently. Removes shared server state (gdsInstalled, one-shot verification flags) that could leak failures across tenants.

Breaking: list-gds-procedures is always registered and appears in list/tools; runtime GDS auto-detection is removed. Initialize now returns clear errors for connectivity, APOC, or misconfiguration instead of failing silently at startup.

### DIFF
--- a/.changes/unreleased/Major-20260519-112432.yaml
+++ b/.changes/unreleased/Major-20260519-112432.yaml
@@ -1,5 +1,5 @@
 kind: Major
 body: | 
-  The list-gds-procedures tool is no longer automatically included in the response from the list/tools endpoint.
-  The Initialize request now provides clearer feedback or error messages when requirements are missing or there is a misconfiguration.
+  The list-gds-procedures tool is no longer auto-detected and included in list/tools responses; whether it is exposed now depends on server-level and/or per-request configuration settings.
+  The Initialize request may now fail when connectivity cannot be verified, required plugins (such as APOC) are missing, or configuration is invalid, with descriptive error messages.
 time: 2026-05-19T11:24:32.375014+01:00
--- a/.changes/unreleased/Major-20260519-112432.yaml
+++ b/.changes/unreleased/Major-20260519-112432.yaml
@@ -1,0 +1,5 @@
+kind: Major
+body: | 
+  The list-gds-procedures tool is no longer automatically included in the response from the list/tools endpoint.
+  The Initialize request now provides clearer feedback or error messages when requirements are missing or there is a misconfiguration.
+time: 2026-05-19T11:24:32.375014+01:00
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -3,7 +3,7 @@
 version: "3"
 
 tasks:
-  # test utils: spass CLI args as task:test -- -v
+  # test utils: pass CLI args as task:test -- -v
   test:
     cmds:
       - go test ./... {{.CLI_ARGS}}
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -2,22 +2,8 @@
 
 version: "3"
 
-dotenv: [".env.task"]
-
 tasks:
-  build:
-    cmds:
-      - go build -C cmd/neo4j-mcp -o ../../bin/ {{.CLI_ARGS}}
-  run:
-    cmds:
-      - go run ./cmd/neo4j-mcp
-  run:compiled:
-    cmds:
-      - ./bin/neo4j-mcp {{.CLI_ARGS}}
-  generate:
-    cmds:
-      - go generate ./...
-  # pass CLI args as task:test -- -v
+  # test utils: spass CLI args as task:test -- -v
   test:
     cmds:
       - go test ./... {{.CLI_ARGS}}
@@ -33,12 +19,32 @@ tasks:
   test:clean:
     cmds:
       - go clean -testcache
+  # build 'n' runs
+  build:
+    cmds:
+      - go build -C cmd/neo4j-mcp -o ../../bin/ {{.CLI_ARGS}}
+  run:
+    dotenv: [".env.task"]
+    cmds:
+      - go run ./cmd/neo4j-mcp
+  run:compiled:
+    dotenv: [".env.task"]
+    cmds:
+      - ./bin/neo4j-mcp {{.CLI_ARGS}}
+  # generation
+  generate:
+    cmds:
+      - go generate ./...
+  # dev tooling
   lint:
     cmds:
       - golangci-lint run
+  # playground 'n' debugging
   inspect:
+    dotenv: [".env.task"]
     cmds:
       - npx @modelcontextprotocol/inspector go run ./cmd/neo4j-mcp
   interactive-client:
+    dotenv: [".env.task"]
     cmds:
       - go run ./client/... bin/neo4j-mcp
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -414,7 +414,7 @@ func (s *Neo4jMCPServer) configureHooks() *server.Hooks {
 
 	hooks.AddAfterCallTool(s.handleToolCallComplete)
 
-	hooks.AddOnRequestInitialization(func(ctx context.Context, id any, message any) error {
+	hooks.AddOnRequestInitialization(func(ctx context.Context, _ any, _ any) error {
 		slog.Info("Initialize request: verifying requirements...")
 		err := s.verifyRequirements(ctx)
 		if err != nil {
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -83,9 +83,7 @@ func NewNeo4jMCPServer(version string, cfg *config.Config, dbService database.Se
 // Start initializes and starts the MCP server
 func (s *Neo4jMCPServer) Start() error {
 	slog.Info("Registering server tools")
-	if err := s.registerTools(); err != nil {
-		return err
-	}
+	s.registerTools()
 	s.emitServerStartupEvent()
 	switch s.config.TransportMode {
 	case config.TransportModeHTTP:
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -7,14 +7,11 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
-	"log"
 	"log/slog"
 	"net/http"
 	"os"
 	"os/signal"
 	"strings"
-	"sync"
-	"sync/atomic"
 	"syscall"
 	"time"
 
@@ -38,25 +35,21 @@ const (
 
 // Neo4jMCPServer represents the MCP server instance
 type Neo4jMCPServer struct {
-	MCPServer          *server.MCPServer
-	httpServer         *http.Server
-	HTTPServerReady    chan struct{}
-	shutdownChan       chan struct{}
-	config             *config.Config
-	dbService          database.Service
-	version            string
-	anService          analytics.Service
-	gdsInstalled       bool
-	initMu             sync.Mutex
-	connectionVerified atomic.Bool
-	uriResolver        URIResolver
-	driverRegistry     database.DriverRegistry
+	MCPServer       *server.MCPServer
+	httpServer      *http.Server
+	HTTPServerReady chan struct{}
+	shutdownChan    chan struct{}
+	config          *config.Config
+	dbService       database.Service
+	version         string
+	anService       analytics.Service
+	uriResolver     URIResolver
+	driverRegistry  database.DriverRegistry
 }
 
 // NewNeo4jMCPServer creates a new MCP server instance
 // The config parameter is expected to be already validated
 func NewNeo4jMCPServer(version string, cfg *config.Config, dbService database.Service, anService analytics.Service) *Neo4jMCPServer {
-
 	neo4jServer := &Neo4jMCPServer{
 		HTTPServerReady: make(chan struct{}),
 		shutdownChan:    make(chan struct{}),
@@ -64,7 +57,6 @@ func NewNeo4jMCPServer(version string, cfg *config.Config, dbService database.Se
 		dbService:       dbService,
 		version:         version,
 		anService:       anService,
-		gdsInstalled:    false,
 	}
 
 	if cfg.TransportMode == config.TransportModeHTTP {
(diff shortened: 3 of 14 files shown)


internal/server/server_test.go
```diff
@@ -217,30 +211,35 @@ func TestNewNeo4jMCPServer(t *testing.T) {
...
+		go func() {
+			startErr = s.Start()
+		}()
+		if startErr != nil {
+			t.Errorf("error while starting the MCP Server")
+		}
+		inProcessClient, err := client.NewInProcessClient(s.MCPServer)
 		if err != nil {
-			t.Errorf("Start() unexpected error = %v", err)
+			t.Fatalf("Unexpected error during InProcesClient creation, %s", err.Error())
```
risset:
```suggestion
			t.Fatalf("Unexpected error during InProcessClient creation, %s", err.Error())
```

risset (dismissed):

internal/server/server_test.go
```diff
@@ -142,62 +114,84 @@ func TestNewNeo4jMCPServer(t *testing.T) {
...
+		if s == nil {
+			t.Errorf("NewNeo4jMCPServer() expected non-nil server, got nil")
+		}
+		var startErr error
+		go func() {
+			startErr = s.Start()
+		}()
+		if startErr != nil {
+			t.Errorf("error while starting the MCP Server")
+		}
```
risset, line 146: Isn't this a race condition? In cases where the return value of `s.Start()` is non-nil, it's possible that `if startErr != nil` will run first and use the initialised value of `startErr`

keremgocen: yes this looks a bit off to me as well, if we set a value of the err in a goroutine we should either use a channel to get the value of it on the main thread or block it with a waitGroup, not sure what was the idea behind it and the test seems to be passing for some reason 🤔

risset: Hmm I noticed that with this `TestInitializeRequestHook` test, the transport is always STDIO which means `Start()` is nonblocking anyway, so shouldn't need to call it in the goroutine then!

internal/server/server.go
```diff
@@ -445,36 +413,16 @@ func (s *Neo4jMCPServer) configureHooks() *server.Hooks {
...
-			if s.gdsInstalled {
-				s.addGDSTools()
-			}
 
-			s.emitConnectionInitializedEvent(ctx)
-
-			s.connectionVerified.Store(true)
-		})
-	}
+	hooks.AddOnRequestInitialization(func(ctx context.Context, _ any, _ any) error {
```
keremgocen, line 415: is this fired after every Initialize request? I'm curious if it's cached on the first one etc. might be worth checking with a multi-tenant test.

internal/server/server_http_lifecycle_test.go
```diff
@@ -142,62 +141,13 @@ func TestNeo4jMCPServerHTTPMode(t *testing.T) {
...
-		if err != nil {
-			t.Fatalf("error while initialize request: %v", err)
-		}
-		// create new client to verify that at new requests the verifyRequirements is not preformed again.
-		mcpClient2 := createStreamableHTTPClient(uri)
-		_, err = mcpClient2.Initialize(context.Background(), mcp.InitializeRequest{})
-		if err != nil {
+		if err == nil {
 			t.Fatalf("error while initialize request: %v", err)
 		}
```
keremgocen, line 148: why are we removing `Server should not perform duplicate verification calls` test? Is that not the case anymore?

Taskfile.yml
```diff
@@ -2,22 +2,8 @@
...
-    cmds:
-      - go run ./cmd/neo4j-mcp
-  run:compiled:
-    cmds:
-      - ./bin/neo4j-mcp {{.CLI_ARGS}}
-  generate:
-    cmds:
-      - go generate ./...
-  # pass CLI args as task:test -- -v
+  # test utils: spass CLI args as task:test -- -v
```
keremgocen:
```suggestion
  # test utils: pass CLI args as task:test -- -v
```

internal/server/server_test.go
```diff
@@ -142,62 +114,84 @@ func TestNewNeo4jMCPServer(t *testing.T) {
...
+		inProcessClient, err := client.NewInProcessClient(s.MCPServer)
+		if err != nil {
+			t.Fatalf("Unexpected error during InProcesClient creation, %s", err.Error())
+		}
+		_, err = inProcessClient.Initialize(context.Background(), mcp.InitializeRequest{})
+		if err == nil {
+			t.Fatal("Expect error during initialization, when no connection can be established, go nil")
```
keremgocen:
```suggestion
			t.Fatal("Expect error during initialization, when no connection can be established, got nil")
```

internal/server/server_test.go
```diff
@@ -142,62 +114,84 @@ func TestNewNeo4jMCPServer(t *testing.T) {
...
+			t.Errorf("error while starting the MCP Server")
+		}
+		inProcessClient, err := client.NewInProcessClient(s.MCPServer)
 		if err != nil {
-			t.Errorf("Start() unexpected error = %v", err)
+			t.Fatalf("Unexpected error during InProcesClient creation, %s", err.Error())
+		}
+		_, err = inProcessClient.Initialize(context.Background(), mcp.InitializeRequest{})
+		if err == nil {
+			t.Fatal("Expect error during initialization, when unexpected results are returned, go nil")
```
keremgocen:
```suggestion
			t.Fatal("Expect error during initialization, when unexpected results are returned, got nil")
```

internal/server/server_test.go
```diff
@@ -217,30 +211,35 @@ func TestNewNeo4jMCPServer(t *testing.T) {
...
+		_, err = inProcessClient.Initialize(context.Background(), mcp.InitializeRequest{})
+		if err != nil {
+			t.Fatalf("Expect no error during initialization, got: %s", err.Error())
 		}
 	})
 }
 
 func TestNewNeo4jMCPServerEvents(t *testing.T) {
+	t.Skip()
+	// move to initialize tests
```
keremgocen, line 231: this looks incomplete, are they moved already?

keremgocen (changes requested):
- we should add a test with STDIO mode Initialize failure
- we should double check the removed tests to make sure we are not removing any existing/expected logic
- we should also fix the race condition, or add an error more explicitly to force a failure in start_server tests
- verify OnRequestInitialization is called only once as we have a metric attached to it

internal/server/server_test.go
MacondoExpress: 🤦  What a good catch! 
The underline reason for this mistake is that in theory ServeStdio is blocking, so when debugging I temporary moved the first test's start in a temporary go routine. 
In practice is it not blocking because it immediately returns when STDIN is closed, which is the case for go test but not for Delve(debugger).
I'll change it, I think I will keep the go-routine to simulate the real word scenario, but I will use a channel or a wg to avoid this pattern to be misused

internal/server/server_http_lifecycle_test.go
MacondoExpress: VerifyRequirements is now per-tenant, as the requirements of Tenant A may not be fulfilled by Tenant B, and is bound to the initialiaze JSON-RPC request.

internal/server/server_test.go
MacondoExpress: You're right, forgot to change this test, good catch as well!

internal/server/server_test.go
MacondoExpress: Actually, I left it as they where before (without the goroutine)

internal/server/server.go
MacondoExpress: This is wraps every initialise requests, which is every time a MCP client starts. This change is part of the multi-tenant requirements, as before we had a single execution of the verifyRequirements which is not longer possible in a multi-tenant environments, as each tenants and related database have different requirements. Same applies for the ConnectionInitializedEvent events as now it should be triggered for each dbs instead of one as it was for the single-tenant design.

MacondoExpress:
> * we should add a test with STDIO mode Initialize failure
> * we should double check the removed tests to make sure we are not removing any existing/expected logic
> * we should also fix the race condition, or add an error more explicitly to force a failure in start_server tests
> * verify OnRequestInitialization is called only once as we have a metric attached to it

This is done now, 
> we should add a test with STDIO mode Initialize failure

this is present in the `TestInitializeRequestHook` test.

> we should also fix the race condition, or add an error more explicitly to force a failure in start_server tests

done, I removed the goroutine, as per comment, which was as it was before. 

> verify OnRequestInitialization is called only once as we have a metric attached to it

OnRequestInitialization is expected to be called multiple times, as per design, as well as the connection initialised event should be emitted when a new connection between the server and a db is made. 
Now, as per multi-tenant design there is no more a single database connected but many, so there will be multiple connection triggered.

internal/server/tools_register.go
```diff
@@ -17,114 +17,53 @@ import (
 // Note: this read-only filtering relies on the tool annotation "readonly" (ReadOnlyHint). If the annotation
 // is not defined or is set to false, the tool will be added (i.e., only tools with readonly=true are filtered in read-only mode).
 func (s *Neo4jMCPServer) registerTools() error {
-	filteredTools := s.getEnabledTools()
-	s.MCPServer.AddTools(filteredTools...)
+	tools := s.getTools()
+	s.MCPServer.AddTools(tools...)
 	return nil
```
luffy1727: Seems like this function is not really returning an error. But server.go seems to be using checking the error.

MacondoExpress: another good catch, I was really sloppy with this PR, I ask for forgiveness 🤦

luffy1727 (dismissed):

internal/server/server_test.go
```diff
@@ -142,62 +114,75 @@ func TestNewNeo4jMCPServer(t *testing.T) {
...
+			t.Errorf("error while starting the MCP Server")
 		}
-
-		// Stop should work without errors
-		ctx := context.Background()
-		err = s.Stop(ctx)
+		inProcessClient, err := client.NewInProcessClient(s.MCPServer)
 		if err != nil {
-			t.Errorf("Stop() unexpected error = %v", err)
+			t.Fatalf("Unexpected error during InProcesClient creation, %s", err.Error())
```
keremgocen:
```suggestion
			t.Fatalf("Unexpected error during InProcessClient creation, %s", err.Error())
```

internal/server/server_test.go
```diff
@@ -142,62 +114,75 @@ func TestNewNeo4jMCPServer(t *testing.T) {
...
+		if s == nil {
+			t.Fatal("NewNeo4jMCPServer() expected non-nil server, got nil")
+		}
+		err := s.Start()
+		if err != nil {
+			t.Errorf("error while starting the MCP Server")
+		}
+		inProcessClient, err := client.NewInProcessClient(s.MCPServer)
+		if err != nil {
+			t.Fatalf("Unexpected error during InProcesClient creation, %s", err.Error())
```
keremgocen:
```suggestion
			t.Fatalf("Unexpected error during InProcessClient creation, %s", err.Error())
```

internal/server/server_test.go
```diff
@@ -142,62 +114,75 @@ func TestNewNeo4jMCPServer(t *testing.T) {
...
 		}
+
 		err := s.Start()
 		if err != nil {
-			t.Errorf("Start() unexpected error = %v", err)
+			t.Errorf("error while starting the MCP Server")
+		}
+		inProcessClient, err := client.NewInProcessClient(s.MCPServer)
+		if err != nil {
+			t.Fatalf("Unexpected error during InProcesClient creation, %s", err.Error())
```
keremgocen:
```suggestion
			t.Fatalf("Unexpected error during InProcessClient creation, %s", err.Error())
```

internal/server/server_test.go
keremgocen: should we remove this comment?

test/integration/server_test.go
```diff
@@ -104,13 +106,21 @@ func TestServerLifecycle(t *testing.T) {
...
+					}
 					if tc.expectError {
-						if startErr == nil {
+						if err == nil {
 							t.Fatal("expected an error but got nil")
 						}
 					} else {
-						if startErr != nil {
-							t.Fatalf("Start returned an unexpected error: %s", startErr.Error())
+						if err != nil {
```
keremgocen, line 122: the `err` here is from L113 `_, err = client.Initialize` not from the `startErr` but that's what we are printing. And I think this will panic when `err != nil and startErr == nil`

.changes/unreleased/Major-20260519-112432.yaml
```diff
@@ -0,0 +1,5 @@
+kind: Major
+body: | 
+  The list-gds-procedures tool is no longer automatically included in the response from the list/tools endpoint.
```
keremgocen: The changelog message looks inverted. It says `list-gds-procedures tool is no longer automatically included`, but I think the changes we have in this PR are the opposite? They are always included.

internal/server/tools_register.go
```diff
@@ -16,115 +16,53 @@ import (
...
-	for _, t := range tools {
-		if t.category != gdsCategory {
-			nonGDSTools = append(nonGDSTools, t)
+	toolsDefs := s.getAllToolsDefs(deps)
+	serverTools := make([]server.ServerTool, 0, len(toolsDefs))
+	for _, toolDef := range toolsDefs {
+		if s.config.ReadOnly && !toolDef.readonly {
+			continue
 		}
+		serverTools = append(serverTools, server.ServerTool{
```
keremgocen: nit: this is unnecessarily inlining `server.ServerTool{Tool: ..., Handler: ...}` we could just do `serverTools = append(serverTools, toolDef.definition)` since `toolDef.definition` is already of type `server.ServerTool`.

keremgocen (changes requested):
a few more changes requested, some are nit but the potential panic in `startErr.Error()` and changelog update seems important.

.changes/unreleased/Major-20260519-112432.yaml
MacondoExpress: Right, I meant that now is no longer dynamically included, I'll find a better phrasing!

keremgocen (approved):

PR #294, opened 2026-05-28
Support filtering of tools returned in tools/list response

risset:
Adds support for configuring the MCP tools that are enabled on the server.
This is exposed via a new environment variable, `NEO4J_MCP_TOOLS`, and equivalent command line flag: `--neo4j-tools`.
Example: `neo4j-mcp --neo4j-tools 'read-cypher,write-cypher'`
When unset, the full list of tools will be enabled.

### DIFF
--- a/internal/cli/args.go
+++ b/internal/cli/args.go
@@ -128,8 +128,10 @@ func ParseConfigFlags() *Args {
 	neo4jDatabase := flag.String("neo4j-database", "", "Neo4j database name (overrides NEO4J_DATABASE env var)")
 	neo4jReadOnly := flag.String("neo4j-read-only", "", "Enable read-only mode: true or false (overrides NEO4J_READ_ONLY env var)")
 	var neo4jTools *string
-	// allows explicit empty string arguments to be differentiated from unset arguments, which isn't possible with flag.String
 	flag.Func("neo4j-tools", "Define tools available by filtering tools returned in tools/list response", func(s string) error {
+		if s == "" {
+			return fmt.Errorf("cannot be empty; omit the flag to use all tools, or provide a comma-separated list of tools")
+		}
 		neo4jTools = &s
 		return nil
 	})
--- a/internal/cli/args.go
+++ b/internal/cli/args.go
@@ -27,6 +27,7 @@ Options:
   --neo4j-password <PASSWORD>         Database password (overrides environment variable NEO4J_PASSWORD)
   --neo4j-database <DATABASE>         Database name (overrides environment variable NEO4J_DATABASE)
   --neo4j-read-only <BOOLEAN>         Enable read-only mode: true or false (overrides environment variable NEO4J_READ_ONLY)
+  --neo4j-tools <TOOLS>               Define tools available by filtering tools returned in tools/list response
   --neo4j-telemetry <BOOLEAN>         Enable telemetry: true or false (overrides environment variable NEO4J_TELEMETRY)
   --neo4j-schema-sample-size <INT>    Number of nodes to sample for schema inference (overrides environment variable NEO4J_SCHEMA_SAMPLE_SIZE)
   --neo4j-transport-mode <MODE>       MCP Transport mode (e.g., 'stdio', 'http') (overrides environment variable NEO4J_TRANSPORT_MODE & NEO4J_MCP_TRANSPORT(deprecated))
@@ -51,6 +52,7 @@ Optional Environment Variables:
   NEO4J_READ_ONLY Enable read-only mode (default: false)
   NEO4J_SCHEMA_SAMPLE_SIZE Number of nodes to sample for schema inference (default: 100)
   NEO4J_TRANSPORT_MODE MCP Transport mode (e.g., 'stdio', 'http') (default: stdio)
+  NEO4J_MCP_TOOLS Define tools available by filtering tools returned in tools/list response (default: all tools enabled)
   NEO4J_MCP_TRANSPORT MCP Transport mode (e.g., 'stdio', 'http') (default: stdio)
   NEO4J_MCP_HTTP_PORT HTTP server port (default: 443 with TLS, 80 without TLS)
   NEO4J_MCP_HTTP_HOST HTTP server host (default: 127.0.0.1)
@@ -79,6 +81,7 @@ type Args struct {
 	Password                          string // #nosec G117 -- Password is only used during startup to create auth token, not logged or exposed
 	Database                          string
 	ReadOnly                          string
+	Tools                             *string // allows explicit empty string arguments to be differentiated from unset arguments
 	Telemetry                         string
 	SchemaSampleSize                  string
 	TransportMode                     string
@@ -101,6 +104,7 @@ var argsSlice = []string{
 	"--neo4j-password",
 	"--neo4j-database",
 	"--neo4j-read-only",
+	"--neo4j-tools",
 	"--neo4j-telemetry",
 	"--neo4j-schema-sample-size",
 	"--neo4j-transport-mode",
@@ -123,6 +127,12 @@ func ParseConfigFlags() *Args {
 	neo4jPassword := flag.String("neo4j-password", "", "Neo4j password (overrides NEO4J_PASSWORD env var)")
 	neo4jDatabase := flag.String("neo4j-database", "", "Neo4j database name (overrides NEO4J_DATABASE env var)")
 	neo4jReadOnly := flag.String("neo4j-read-only", "", "Enable read-only mode: true or false (overrides NEO4J_READ_ONLY env var)")
+	var neo4jTools *string
+	// allows explicit empty string arguments to be differentiated from unset arguments, which isn't possible with flag.String
+	flag.Func("neo4j-tools", "Define tools available by filtering tools returned in tools/list response", func(s string) error {
+		neo4jTools = &s
+		return nil
+	})
 	neo4jTelemetry := flag.String("neo4j-telemetry", "", "Enable telemetry: true or false (overrides NEO4J_TELEMETRY env var)")
 	neo4jSchemaSampleSize := flag.String("neo4j-schema-sample-size", "", "Number of nodes to sample for schema inference (overrides NEO4J_SCHEMA_SAMPLE_SIZE env var)")
 	neo4jTransportMode := flag.String("neo4j-transport-mode", "", "MCP Transport mode (e.g., 'stdio', 'http') (overrides NEO4J_TRANSPORT_MODE env var)")
@@ -144,6 +154,7 @@ func ParseConfigFlags() *Args {
 		Password:                          *neo4jPassword,
 		Database:                          *neo4jDatabase,
 		ReadOnly:                          *neo4jReadOnly,
+		Tools:                             neo4jTools,
 		Telemetry:                         *neo4jTelemetry,
 		SchemaSampleSize:                  *neo4jSchemaSampleSize,
 		TransportMode:                     *neo4jTransportMode,
--- a/internal/cli/args_test.go
+++ b/internal/cli/args_test.go
@@ -138,6 +138,24 @@ func TestHandleArgs(t *testing.T) {
 			expectedExitCode: -1, // Should not exit, flag is allowed
 		},
 		{
+			name:             "neo4j-tools flag with one tool name",
+			args:             []string{testProgramName, "--neo4j-tools", "get-schema"},
+			version:          testVersion,
+			expectedExitCode: -1, // Should not exit, flag is allowed
+		},
+		{
+			name:             "neo4j-tools flag with trailing comma after tool name",
+			args:             []string{testProgramName, "--neo4j-tools", "get-schema,"},
+			version:          testVersion,
+			expectedExitCode: -1, // Should not exit, flag is allowed
+		},
+		{
+			name:             "neo4j-tools flag with two tool names",
+			args:             []string{testProgramName, "--neo4j-tools", "get-schema,read-cypher"},
+			version:          testVersion,
+			expectedExitCode: -1, // Should not exit, flag is allowed
+		},
+		{
 			name:             "multiple configuration flags",
 			args:             []string{testProgramName, "--neo4j-uri", "bolt://localhost:7687", "--neo4j-username", "user"},
 			version:          testVersion,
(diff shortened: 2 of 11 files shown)


internal/config/config_test.go
```diff
@@ -912,3 +925,126 @@ func TestLoadConfig_HTTPModeDatabase(t *testing.T) {
...
+			cliTools: ",,",
+			wantErr:  `tool "" is invalid. Available tools are: read-cypher, write-cypher, list-gds-procedures, get-schema`,
+		},
+		// TODO: Not implement yet
+		// {
+		// 	name:          "When tools is provided as empty string, it should disable all tools",
+		// 	cliTools:      "",
+		// 	expectedTools: []string{},
+		// },
+	}
```
risset, line 1054: We had a discussion about this when pairing, and our intended behaviour was that `--neo4j-tools ""` should be interpreted as an explicit disabling of all tools,  <strike>but we're still working on implementing that correctly</strike>. This has been implemented now

internal/server/server.go
```diff
@@ -82,7 +82,6 @@ func NewNeo4jMCPServer(version string, cfg *config.Config, dbService database.Se
 
 // Start initializes and starts the MCP server
 func (s *Neo4jMCPServer) Start() error {
-	slog.Info("Registering server tools")
```
risset, line 85: Removed since the logging now happens where tools are registered, which includes the list and number of enabled tools as fields

internal/cli/args.go
```diff
@@ -27,6 +27,7 @@ Options:
   --neo4j-password <PASSWORD>         Database password (overrides environment variable NEO4J_PASSWORD)
   --neo4j-database <DATABASE>         Database name (overrides environment variable NEO4J_DATABASE)
   --neo4j-read-only <BOOLEAN>         Enable read-only mode: true or false (overrides environment variable NEO4J_READ_ONLY)
+  --neo4j-tools <[]STRING>            Define tools available by filtering tools returned in tools/list response
```
keremgocen: just exploring the thought of converting these into something more practical and easy to use, as they are project-scoped and unlikely to leak out or create conflict like the env vars we have. How do we feel about making this `--tools`, *possibly `-t` for short*, instead of `--neo4j-tools`? Do we need `neo4j` in there for the flags at all? (it's a nit, and should be part of the env flag overhaul ticket we have in linear and not to be done within the scope of this PR, just want to ignite the discussion on it)

internal/server/tool_register_test.go
```diff
@@ -109,5 +114,54 @@ func TestToolRegister(t *testing.T) {
 			t.Errorf("Expected %d tools, but test configuration shows %d", expectedTotalToolsCount, registeredTools)
 		}
 	})
+	t.Run("should only register tools that are specified in config", func(t *testing.T) {
```
keremgocen, line 118: 💯

internal/server/tools_register.go
```diff
@@ -37,6 +46,9 @@ func (s *Neo4jMCPServer) getTools() []server.ServerTool {
 		if s.config.ReadOnly && !toolDef.readonly {
 			continue
 		}
+		if !slices.Contains(s.config.Tools, toolDef.definition.Tool.Name) {
```
keremgocen: nit: why do we have this bit? Is it always guaranteed to have a populated `toolDef.definition.Tool.Name` value?

internal/cli/args_test.go
```diff
@@ -137,6 +137,24 @@ func TestHandleArgs(t *testing.T) {
...
 			expectedExitCode: -1, // Should not exit, flag is allowed
 		},
+		{
+			name:             "neo4j-tools configuration flag",
+			args:             []string{testProgramName, "--neo4j-tools", "get-schema"},
+			version:          testVersion,
+			expectedExitCode: -1, // Should not exit, flag is allowed
+		},
+		{
+			name:             "neo4j-tools configuration flag",
```
keremgocen: Can we please use separate test names for these 3 tests so we can distinguish between them in the output?
`e.g. neo4j-tools single tool, ..trailing comma, ..multiple tools`

internal/server/tools_register.go
keremgocen: Let's add a test to ensure the names in `AvailableTools` match `getAllToolsDefs`. The problem with the current version is, if someone adds a tool and forgets to update `AvailableTools`, the new tool can never be enabled.

luffy1727:
Can we maybe tidy up and squash the commits. 14 seems like a lot and there is a WIP commit in it as well

internal/cli/args.go
risset: Yeah I don't see much benefit to the `neo4j` prefix

risset:
> Can we maybe tidy up and squash the commits. 14 seems like a lot and there is a WIP commit in it as well

Yeah that makes sense and happy to do this.
_Edit_: can also squash merge, when merging, I wasn't sure what the team's preferences were in this regard. 
I do also prefer to keep the branch tidy especially if the PR is ongoing for a while like in this case

internal/server/tool_register_test.go
```diff
@@ -109,5 +114,54 @@ func TestToolRegister(t *testing.T) {
...
+		}
+		s := server.NewNeo4jMCPServer("test-version", cfg, mockDB, aService)
+
+		// Expected tools that should be registered
+		expectedTools := []string{"get-schema", "read-cypher"}
+
+		// Start server and register tools
+		err := s.Start()
+		if err != nil {
+			t.Fatalf("Start() failed: %v", err)
```
keremgocen: can we consider `require.NoError(t, err)` to keep things consistent?

internal/server/tools_register.go
```diff
@@ -19,6 +22,12 @@ import (
 func (s *Neo4jMCPServer) registerTools() {
```
keremgocen, line 22: The comment for this function (`registerTools`) is a bit stale, and probably unnecessarily long now. Can we please update it to include the new tools filtering logic as well?

internal/server/tools_register.go
```diff
@@ -37,6 +46,9 @@ func (s *Neo4jMCPServer) getTools() []server.ServerTool {
 		if s.config.ReadOnly && !toolDef.readonly {
```
keremgocen: I think we should also add a warning here, because if ReadOnly=true now it will filter out expected tools silently in the config if the tool filter is also present. It could help debuggers if we warn the observers with something along the lines of `"tool skipped: excluded by read-only mode"`.

internal/cli/args.go
keremgocen: another nit: `[]STRING` is not very helpful here as a help text, using something like `<tool1,tool2,...>` or `<TOOLS>` would be better I think.

MacondoExpress: Yeah, I agree I think we have a card to remove these as well to make the environment variable consistent.
I would suggest to do this change in a single PR with all the deprecation while keeping this PR consistent with the previous approach, wdyt ?
We're happy both way, however if we change it now, then it will be inconsistent with others in case we're going 2.x GA without doing that card

internal/config/config_test.go
keremgocen: > our intended behaviour was that --neo4j-tools "" should be interpreted as an explicit disabling of all tools

I'm not sure about this. It semantically conflicts a bit in my head with the golang string taking "" as the default value. As a user I think it's more explicit to set it to a specific value such as `None` instead of leaving it empty to disable everything. Why would we want to disable all tools anyway?

internal/server/tools_register.go
MacondoExpress: Can you please expand this comment? we're not fully understand it.

internal/config/config.go
```diff
@@ -240,6 +259,12 @@ func LoadConfig(cliOverrides *CLIOverrides) (*Config, error) {
 		}
 	}
 
+	// If tools haven't been set at this point, they have neither been provided nor explicitly unset
+	// Default to all available tools
+	if cfg.Tools == nil {
+		cfg.Tools = AvailableTools
```
keremgocen, line 267: my final teeny-tiny nit, and possibly falling a bit on the overly defensive side, instead of a review item, should we give the config its own clone (`slices.Clone(AvailableTools)`) as otherwise they'll be sharing the same backing array. In the unlikely scenario of cfg.Tools being modified outside the package, it will corrupt the shared slice. (claude's review tool caught this to be fair, and I think it's low prio or a non-issue, so feel free to ignore)

keremgocen (dismissed):
I don't have anything blocking, so I'm tentatively approving, left some comments. Nice work! 🚀 

List of my comments worth addressing before merging:
- Duplicate test names in args_test.go — three cases all named "neo4j-tools configuration flag"
- `AvailableTools` / `getAllToolsDefs` coupling — no test ensuring the string names stay in sync
- `require.NoError` vs `t.Fatalf` in the new tool_register_test.go tests

internal/config/config.go
risset: Yeah I also did a code review with Claude Code and it raised this for me too.

I didn't implement it because it didn't seem that likely we would be affected by this, and it makes the code a bit more obscure to readers that might not be aware of this slice behaviour...

but now that you have also mentioned it, I think we can make this change and add a comment, since it can be helpful to be aware of this as well for other places it might be more important

internal/config/config_test.go
risset: > Why would we want to disable all tools anyway

I think, one reason is that potentially the server could serve not only tools in the future, but prompts or resources. In which case, it might be valid to disable tools while allowing those. After typing that out though, I realise maybe it wouldn't be useful anyway to have other primitives without any tools...

> As a user I think it's more explicit to set it to a specific value such as None instead of leaving it empty to disable everything.

Hmm, I get that although I did think `--neo4j-tools ""` seems to make more sense than having a keyword which could be mixed up with the tool names themselves.

The "" being default string value in Go really does make this behaviour annoying to handle yeah, hence ending up with a solution involving `*string` and `flag.Func`, which feels like a workaround. So I wouldn't be opposed to just not supporting this behaviour anyway.

internal/server/tool_register_test.go
risset: I think this makes sense, though I found 5 instances of `t.Fatalf("Start() failed: %v", err)` in the file, so perhaps it would be better to change them in another PR since this one is already quite big

internal/server/tools_register.go
keremgocen: I made this comment in my initial pass, now I see that this is the actual filtering logic, and we silently ignore tools not in the list, which is ok, I think, but maybe needs a more explicit `warn` since readOnly can cause a different behaviour than expected by the user.

The other thing I was wondering is if we could ever get a panic by calling `toolDef.definition.Tool.Name` and if anything in between could be nil at some point, somehow. Just in case I'm missing something.

internal/server/tool_register_test.go
risset: We can change it for just this added code, in this PR for now though

internal/config/config_test.go
keremgocen: >it might be valid to disable tools while allowing those

After reading this I'm leaning even more towards not having it:) We are currently not supporting prompts or resources though and it feels a bit of an overkill to do anything along the lines of "we might need this in the future". I think for now we just don't need this functionality. And we can always add it later without a breaking change, removing it later however, would be a breaking change.

MacondoExpress: We could raise an error in case of an empty string,
In that way, we avoid complicating the codebase, and we don't support "disable all tools" from the start, as it does not make sense in the current state of the MCP server. 
Then, if/when this behaviour is needed, we can remove the error and support an explicit empty string as an additive.

keremgocen: Yeah I guess returning an error still sounds better than implementing features we don't need yet.

internal/cli/args_test.go
```diff
@@ -137,6 +137,24 @@ func TestHandleArgs(t *testing.T) {
...
+			args:             []string{testProgramName, "--neo4j-tools", "get-schema,"},
+			version:          testVersion,
+			expectedExitCode: -1, // Should not exit, flag is allowed
+		},
+		{
+			name:             "neo4j-tools flag with two tool names",
+			args:             []string{testProgramName, "--neo4j-tools", "get-schema,read-cypher"},
+			version:          testVersion,
+			expectedExitCode: -1, // Should not exit, flag is allowed
+		},
```
MacondoExpress, line 157: here is missing the error case, we just added

internal/config/config.go
```diff
@@ -178,6 +189,10 @@ func LoadConfig(cliOverrides *CLIOverrides) (*Config, error) {
 		AllowUnauthenticatedToolsList: ParseBool(GetEnv("NEO4J_HTTP_ALLOW_UNAUTHENTICATED_TOOLS_LIST"), false),
 	}
 
+	if toolsEnv := GetEnv("NEO4J_MCP_TOOLS"); toolsEnv != "" {
```
MacondoExpress: in my opinion `NEO4J_MCP_TOOL`="" and --`neo4j-tools`="" should return the same error and mimic each other behaviour

MacondoExpress (changes requested):

internal/config/config.go
risset: > in my opinion `NEO4J_MCP_TOOL`="" and --`neo4j-tools`="" should return the same error and mimic each other behaviour

I agree, but then I thought that `NEO4J_MCP_TOOL=""` combined with a valid value for the `neo4j-tools` flag, also shouldn't be an error

internal/cli/args_test.go
risset: It doesn't seem like we can assert against --neo4j-tools "" exiting here.
We don't return the error until ParseConfigFlags which is after HandleArgs has been called, and we don't have unit test coverage for it

internal/config/config.go
risset: If we want to be strict with enforcing the env var is valid regardless we can do that though

internal/cli/args_test.go
MacondoExpress: ah you're right, I assumed we were testing that! Sounds good then!

internal/config/config.go
MacondoExpress: I think it make more sense to validate it and treat the same as for the arg, I'll expand the thought here:
- Right now all the args behaviour mimic the env, this make user's expectation clear and it does not require us to document exception in our documentation
- It will allow to use explicit empty quote in the future, without relying on breaking change, allowing for a simple migration path

MacondoExpress (approved):

keremgocen (approved):
LGTM!

PR #297, opened 2026-06-01
Per-request tool filtering via X-Neo4j-MCP-Tools and X-Neo4j-MCP-Readonly

MacondoExpress:
Adds two new HTTP request header:  X-Neo4j-MCP-Readonly and X-Neo4j-MCP-Tools,  that let clients dynamically filter the tools advertised by the MCP server on a per-request basis, without restarting the server.

- `X-Neo4j-MCP-Readonly`: "true/false" restricts the tool list to read-only tools (using the MCP spec's readOnlyHint as the source of truth)
- `X-Neo4j-MCP-Tools`: "read-cypher, get-schema" restricts the tool list to an explicit comma-separated subset; invalid tool names return 400

When both headers are set, they are applied as an intersection, this means that even if explicitly set by `X-Neo4j-MCP-Tools` tools that are not readonly cannot be enabled in readonly mode.
When header are not set the behaviour remains the same as before.

Refactors tools_register.go to derive read-only classification from ReadOnlyHint annotations rather than a separate readonly field on ToolDefinition


internal/server/tools_register.go
```diff
@@ -42,49 +38,40 @@ func (s *Neo4jMCPServer) getTools() []server.ServerTool {
...
-		serverTools = append(serverTools, toolDef.definition)
+		serverTools = append(serverTools, toolDef)
 	}
 	return serverTools
 }
 
 // getAllToolsDefs returns all available tools with their specs and handlers
-func (s *Neo4jMCPServer) getAllToolsDefs(deps *tools.ToolDependencies) []ToolDefinition {
-	return []ToolDefinition{
+func (s *Neo4jMCPServer) getAllToolsDefs(deps *tools.ToolDependencies) []server.ServerTool {
```
risset: I think the name of this function is outdated now that we're returning `[]server.ServerTool` instead of `[]ToolDefinition`...

Would it be better to just inline the `[]server.ServerTool` in `getTools()` instead of having this extra function?

internal/server/server.go
```diff
@@ -73,6 +74,22 @@ func NewNeo4jMCPServer(version string, cfg *config.Config, dbService database.Se
 		server.WithHooks(hooks),
 		server.WithInstructions("This is the Neo4j official MCP server and can provide tool calling to interact with your Neo4j database,"+
 			"by inferring the schema with tools like get-schema and executing arbitrary Cypher queries with read-cypher."),
+		server.WithToolFilter(func(ctx context.Context, tools []mcp.Tool) []mcp.Tool {
+			// This hook filter tools depending on the X-Neoj4-MCP-Tools and X-Neo4j-MCP-Readonly http header
```
risset:
```suggestion
			// This hook filter tools depending on the X-Neo4j-MCP-Tools and X-Neo4j-MCP-Readonly http header
```

internal/server/server.go
```diff
@@ -73,6 +74,22 @@ func NewNeo4jMCPServer(version string, cfg *config.Config, dbService database.Se
...
+			readonly := mcpcontext.GetReadonly(ctx)
+			if readonly == nil || *readonly == false {
+				return tools
+			}
+			var readonlyTools = make([]mcp.Tool, 0, 4)
```
risset: Can we use `len(tools)` here instead of hard-coding `4`?

risset (changes requested):

internal/server/tools_register.go
MacondoExpress: Yeah, still in draft however, not ready to be reviewed. It still has to be polished

MacondoExpress: Is your preference inline btw?

risset: > Is your preference inline btw?

Hmm I prefer it slightly but not a big deal :)

internal/mcpcontext/mcpcontext.go
```diff
@@ -69,3 +71,37 @@ func GetDriver(ctx context.Context) (any, bool) {
...
+	readonly, ok := ctx.Value(readonlyKey).(bool)
+	if !ok {
+		// The assumption here is that ctx entry is set by the setter defined in mcpcontext,
+		// therefore is nil when ok is false
+		return nil
+	}
+	return &readonly
+}
+
+// WithTools store per-request tools in the contex
```
risset:
```suggestion
// WithTools store per-request tools in the context
```

internal/server/middleware.go
```diff
@@ -252,6 +256,76 @@ func pathValidationMiddleware() func(http.Handler) http.Handler {
...
+// "false" (case-insensitive). Any other non-empty value yields a 400 Bad Request.
+// When the header is absent it does not set the Readonly in the context.
+func readonlyMiddleware() func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// r.Header.Values is been used to distinguish when the header is not set
+			vals := r.Header.Values("X-Neo4j-MCP-Readonly")
+
+			if len(vals) > 1 {
+				http.Error(w, "Bad Request: Ambiguity X-Neo4j-MCP-Readonly header found", http.StatusBadRequest)
```
risset:
```suggestion
				http.Error(w, "Bad Request: Ambiguous X-Neo4j-MCP-Readonly header found", http.StatusBadRequest)
```

internal/server/middleware_test.go
```diff
@@ -869,3 +855,126 @@ func TestNeo4jDriverMiddleware_ErrorPaths(t *testing.T) {
...
+				assert.Equal(t, tt.wantReadonly, *gotReadonly)
+
+			}
+			if !tt.wantReadonlySet {
+				assert.Nil(t, gotReadonly)
+			}
+
+		})
+	}
+}
```
risset, line 965: Does `toolsMiddleware` also need tests?

risset (changes requested):

internal/server/server.go
```diff
@@ -73,6 +75,27 @@ func NewNeo4jMCPServer(version string, cfg *config.Config, dbService database.Se
...
+			// This hook filter tools depending on the X-Neo4j-MCP-Tools and X-Neo4j-MCP-Readonly http header
+			readonly := mcpcontext.GetReadonly(ctx)
+			requestedTools := mcpcontext.GetTools(ctx)
+			// early return when no per-request filters are not defined
+			if readonly == nil && requestedTools == nil {
+				return tools
+			}
+			var filteredTools = make([]mcp.Tool, 0, len(tools))
+			for _, tool := range tools {
+				if readonly != nil && *readonly && !*tool.Annotations.ReadOnlyHint {
```
risset: Is there a case where `tool.Annotations.ReadOnlyHint` can be nil here?
I guess it's only if we forgot to add `WithReadOnlyHintAnnotation` in the tool spec files

MacondoExpress: exactly this, only if is not set on the tool spec

internal/server/middleware_test.go
MacondoExpress: right, forgot to add it, brb

internal/server/server.go
```diff
@@ -73,6 +75,27 @@ func NewNeo4jMCPServer(version string, cfg *config.Config, dbService database.Se
...
+			// This hook filter tools depending on the X-Neo4j-MCP-Tools and X-Neo4j-MCP-ReadOnly http header
+			readOnly := mcpcontext.GetReadOnly(ctx)
+			requestedTools := mcpcontext.GetTools(ctx)
+			// early return when no per-request filters are not defined
+			if readOnly == nil && requestedTools == nil {
+				return tools
+			}
+			var filteredTools = make([]mcp.Tool, 0, len(tools))
+			for _, tool := range tools {
+				if readOnly != nil && *readOnly && !*tool.Annotations.ReadOnlyHint {
```
keremgocen: I think this is a bug as we can't rely on ReadOnlyHint being set which is happening somewhere else.
We already have similar nil guard in tools_register.go so we should do the same here: `(tool.Annotations.ReadOnlyHint == nil || !*tool.Annotations.ReadOnlyHint)`

internal/mcpcontext/mcpcontext.go
```diff
@@ -69,3 +71,37 @@ func GetDriver(ctx context.Context) (any, bool) {
...
+// Returns nil if the tools are not explicitly requested.
+func GetTools(ctx context.Context) *[]string {
+	tools, ok := ctx.Value(toolsKey).([]string)
+	if !ok {
+		// The assumption here is that ctx entry is set by the setter defined in mcpcontext,
+		// therefore is nil when ok is false
+		return nil
+	}
+	return &tools
+}
```
keremgocen: nit: missing new line (not sure why this was not picked up by the editor)
```suggestion
}

```

internal/server/middleware.go
```diff
@@ -252,6 +256,76 @@ func pathValidationMiddleware() func(http.Handler) http.Handler {
...
+			vals := r.Header.Values("X-Neo4j-MCP-Tools")
+			if len(vals) > 1 {
+				http.Error(w, "Bad Request: Ambiguity X-Neo4j-MCP-Tools header found", http.StatusBadRequest)
+				return
+			} else if len(vals) == 1 {
+				if vals[0] == "" {
+					http.Error(w, fmt.Sprintf("tool \"\" is invalid. Available tools are: %s", strings.Join(config.AvailableTools, ", ")), http.StatusBadRequest)
+					return
+				}
+				tools := parseCommaSeparatedString(vals[0])
```
keremgocen: I think we have another bug here, `X-Neo4j-MCP-Tools: ,,,` would pass the empty string check which we normally return 400 I think. We need to check the tools length after doing parseCommaSeparatedString:
`if len(tools) == 0 {`...

internal/server/tools_register.go
```diff
@@ -30,63 +30,49 @@ func (s *Neo4jMCPServer) registerTools() {
...
+			Tool:    cypher.GetSchemaSpec(),
+			Handler: cypher.GetSchemaHandler(deps, s.config.SchemaSampleSize),
 		},
 		{
-			definition: server.ServerTool{
-				Tool:    cypher.ReadCypherSpec(),
-				Handler: cypher.ReadCypherHandler(deps),
-			},
-			readonly: true,
+
```
keremgocen, line 62:
```suggestion
```

internal/server/tools_register.go
```diff
@@ -30,63 +30,49 @@ func (s *Neo4jMCPServer) registerTools() {
...
+			Tool:    cypher.ReadCypherSpec(),
+			Handler: cypher.ReadCypherHandler(deps),
 		},
 		{
-			definition: server.ServerTool{
-				Tool:    cypher.WriteCypherSpec(),
-				Handler: cypher.WriteCypherHandler(deps),
-			},
-			readonly: false,
+
```
keremgocen, line 67:
```suggestion
```

internal/server/tools_register.go
```diff
@@ -30,63 +30,49 @@ func (s *Neo4jMCPServer) registerTools() {
...
+			Handler: cypher.WriteCypherHandler(deps),
 		},
 		// GDS Category/Section
 		{
-			definition: server.ServerTool{
-				Tool:    gds.ListGDSProceduresSpec(),
-				Handler: gds.ListGdsProceduresHandler(deps),
-			},
-			readonly: true,
+
```
keremgocen, line 73:
```suggestion
```
nit: some stray empty lines here

internal/server/middleware_test.go
```diff
@@ -869,3 +855,235 @@ func TestNeo4jDriverMiddleware_ErrorPaths(t *testing.T) {
...
+
+			handler.ServeHTTP(rec, req)
+
+			assert.Equal(t, tt.wantCode, rec.Code)
+
+			if tt.wantReadOnlySet {
+				assert.Equal(t, tt.wantReadOnly, *gotReadOnly)
+
+			}
+			if !tt.wantReadOnlySet {
```
keremgocen:
```suggestion
			else {
```
nit: this could just negate the previous if block

internal/server/middleware.go
```diff
@@ -252,6 +256,76 @@ func pathValidationMiddleware() func(http.Handler) http.Handler {
...
+
+// toolsMiddleware reads the "X-Neo4j-MCP-Tools" request header and stores
+// the resulting values in the request context. Accepted values are the tools
+// supported by the Neo4j MCP Server such as: "read-cypher", "get-schema".
+// Any other non-empty value yields a 400 Bad Request.
+// When the header is absent it does not set the Tools in the context.
+func toolsMiddleware() func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// r.Header.Values is been used to distinguish when the header is not set
```
keremgocen:
```suggestion
			// r.Header.Values is used to distinguish when the header is not set
```
more nits:)

internal/server/middleware.go
```diff
@@ -252,6 +256,76 @@ func pathValidationMiddleware() func(http.Handler) http.Handler {
...
 }
 
+// readOnlyMiddleware reads the "X-Neo4j-MCP-ReadOnly" request header and stores
+// the resulting boolean in the request context. Accepted values are "true" and
+// "false" (case-insensitive). Any other non-empty value yields a 400 Bad Request.
+// When the header is absent it does not set the ReadOnly in the context.
+func readOnlyMiddleware() func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// r.Header.Values is been used to distinguish when the header is not set
```
keremgocen:
```suggestion
			// r.Header.Values is used to distinguish when the header is not set
```

internal/server/middleware.go
```diff
@@ -252,6 +256,76 @@ func pathValidationMiddleware() func(http.Handler) http.Handler {
...
+// supported by the Neo4j MCP Server such as: "read-cypher", "get-schema".
+// Any other non-empty value yields a 400 Bad Request.
+// When the header is absent it does not set the Tools in the context.
+func toolsMiddleware() func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// r.Header.Values is been used to distinguish when the header is not set
+			vals := r.Header.Values("X-Neo4j-MCP-Tools")
+			if len(vals) > 1 {
+				http.Error(w, "Bad Request: Ambiguity X-Neo4j-MCP-Tools header found", http.StatusBadRequest)
```
keremgocen:
```suggestion
				http.Error(w, "Bad Request: duplicate X-Neo4j-MCP-Readonly header", http.StatusBadRequest)
```
maybe?

keremgocen (changes requested):
nice one! 
Mostly looking good to me except one potential runtime panic in `WithToolFilter`, the rest are cosmetic updates.

internal/server/server.go
MacondoExpress: I agree no reason to keep it as is it. Not really a bug anyway as we're relying on the Readonly hint to be set in the project, so this hypothetical condition is not reachable. However, no reason to fail badly here in case a conformant tool will be mistakenly added. Let me check if I can easily enforce the readonly in the tools as well!

internal/server/middleware.go
MacondoExpress: hmm nice catch!

internal/server/tools_register.go
MacondoExpress: I resolved this as I don't have found any empty lines

internal/mcpcontext/mcpcontext.go
```diff
@@ -69,3 +71,37 @@ func GetDriver(ctx context.Context) (any, bool) {
 	driver := ctx.Value(driverKey)
 	return driver, driver != nil
 }
+
+// WithReadOnly marks the context as read-only
+func WithReadOnly(ctx context.Context, readOnly bool) context.Context {
+	return context.WithValue(ctx, readOnlyKey, readOnly)
+}
+
+// GetReadOnly retrieves the read-only flag from the context.
```
luffy1727, line 80: I am not really sure what these comments accomplish to be honest. I feel like, these are basically just explaining what the code does and increasing surface with very little value

internal/server/middleware.go
```diff
@@ -252,6 +256,76 @@ func pathValidationMiddleware() func(http.Handler) http.Handler {
...
+func readOnlyMiddleware() func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// r.Header.Values is used to distinguish when the header is not set
+			vals := r.Header.Values("X-Neo4j-MCP-ReadOnly")
+
+			if len(vals) > 1 {
+				http.Error(w, "Bad Request: duplicate X-Neo4j-MCP-ReadOnly header found", http.StatusBadRequest)
+				return
+			} else if len(vals) == 1 {
```
luffy1727: You don't need the else statement here. Just regular `if` should do the trick

internal/server/middleware.go
```diff
@@ -252,6 +256,76 @@ func pathValidationMiddleware() func(http.Handler) http.Handler {
...
+				switch strings.ToLower(vals[0]) {
+				case "false":
+					ctx := mcpcontext.WithReadOnly(r.Context(), false)
+					next.ServeHTTP(w, r.WithContext(ctx))
+					return
```
luffy1727: can every return be simplified with 1 return in the end

internal/server/middleware.go
```diff
@@ -252,6 +256,76 @@ func pathValidationMiddleware() func(http.Handler) http.Handler {
...
+// When the header is absent it does not set the Tools in the context.
+func toolsMiddleware() func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// r.Header.Values is used to distinguish when the header is not set
+			vals := r.Header.Values("X-Neo4j-MCP-Tools")
+			if len(vals) > 1 {
+				http.Error(w, "Bad Request: duplicate X-Neo4j-MCP-Tools header found", http.StatusBadRequest)
+				return
+			} else if len(vals) == 1 {
```
luffy1727: Here as well you don't need the else statement

internal/server/middleware.go
```diff
@@ -252,6 +256,76 @@ func pathValidationMiddleware() func(http.Handler) http.Handler {
...
+// supported by the Neo4j MCP Server such as: "read-cypher", "get-schema".
+// Any other non-empty value yields a 400 Bad Request.
+// When the header is absent it does not set the Tools in the context.
+func toolsMiddleware() func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// r.Header.Values is used to distinguish when the header is not set
+			vals := r.Header.Values("X-Neo4j-MCP-Tools")
+			if len(vals) > 1 {
+				http.Error(w, "Bad Request: duplicate X-Neo4j-MCP-Tools header found", http.StatusBadRequest)
```
luffy1727, line 304: what do we think about creating a error object or string that we can re-use. It seems like most of the errors are duplicated with almost identical messages

keremgocen (dismissed):
Thanks for addressing the comments, lgtm now!

internal/server/middleware.go
MacondoExpress: we can, but I'm not sure if it will be more readable in that way

internal/server/middleware.go
MacondoExpress: You mean the one below:`tool %q is invalid. Available tools are: %s` ? this one is as an unique message

internal/mcpcontext/mcpcontext.go
MacondoExpress: it's just consistent with the rest of the file, not that I completely disagree tho

luffy1727 (approved):

risset (approved):

PR #298, opened 2026-06-06
fix: declare explicit JSON Schema for tool inputs

kerandia:
## Summary

The `tools/list` response currently emits malformed JSON Schemas for four tools because the input-schema generation in `mcp-go` silently fails on the input types used:

- **`read-cypher` / `write-cypher`** use `mcp.WithInputSchema[ReadCypherInput]()`, where `ReadCypherInput.Params` is a custom map type (`Params`) with `UnmarshalJSON`. The `google/jsonschema-go` reflector can't derive a schema for it, so the emitted schema has `properties: {}` — no declared parameters at all, even though the tool requires `query`.
- **`get-schema` and `list-gds-procedures`** use `mcp.WithInputSchema[EmptyInput]()` with a struct that has a `Properties` field. The emitted schema requires a parameter literally named `"properties"`, which is nonsense for tools that take no arguments.

Forgiving clients (e.g. claude.ai) ignore these schemas; stricter MCP clients reject the tools outright. The tool descriptions themselves are correct — only the parameter declaration is wrong.

## What this PR does

Declares the input schemas explicitly with `WithRawInputSchema` via a small `rawInputSchema` helper:

```go
func rawInputSchema(schema json.RawMessage) mcp.ToolOption {
    return func(t *mcp.Tool) {
        t.InputSchema = mcp.ToolInputSchema{}
        t.RawInputSchema = schema
    }
}
```

The helper clears the default `InputSchema` that `mcp.NewTool` pre-populates (`Type: \"object\"`) — without that step, `Tool.MarshalJSON` errors with `\"provide either InputSchema or RawInputSchema, not both\"`.

`ReadCypherInput` / `WriteCypherInput` structs are retained because they're used for argument binding in the handlers and in tests; only the unused `jsonschema:` tag was removed. `EmptyInput` is dropped — it was only used as a (broken) schema placeholder.

## Before / after

`tools/list` for `read-cypher` (and `write-cypher`):

```diff
 \"inputSchema\": {
   \"type\": \"object\",
-  \"properties\": {},
-  \"required\": []
+  \"properties\": {
+    \"query\":  { \"type\": \"string\", \"description\": \"The Cypher query to execute\" },
+    \"params\": { \"type\": \"object\", \"description\": \"Parameters to pass to the Cypher query\", \"additionalProperties\": true }
+  },
+  \"required\": [\"query\"],
+  \"additionalProperties\": false
 }
```

`tools/list` for `get-schema` (and `list-gds-procedures`):

```diff
 \"inputSchema\": {
   \"type\": \"object\",
-  \"properties\": {
-    \"properties\": { \"type\": \"object\", \"additionalProperties\": true }
-  },
-  \"required\": [\"properties\"]
+  \"properties\": {},
+  \"additionalProperties\": false
 }
```

## Test plan

- [x] `go test ./internal/tools/cypher/... ./internal/tools/gds/...` passes locally (in the official builder Docker image)
- [x] Built the fixed image, ran the server in HTTP transport mode, and confirmed `tools/list` returns the schemas above
- [x] Confirmed `tools/call` for `read-cypher` with `{\"query\": \"MATCH (p) RETURN count(p)\"}` still works end-to-end against a real Neo4j Aura instance

### DIFF
--- a/internal/tools/cypher/get_schema_spec.go
+++ b/internal/tools/cypher/get_schema_spec.go
@@ -4,22 +4,21 @@
 package cypher
 
 import (
+	"encoding/json"
+
 	"github.com/mark3labs/mcp-go/mcp"
 )
-type EmptyInput struct {
-    Properties map[string]interface{} `json:"properties"`
-}
 
 func GetSchemaSpec() mcp.Tool {
 	return mcp.NewTool("get-schema",
 		mcp.WithDescription(`
 		Retrieve the schema information from the Neo4j database, including node labels, relationship types, and property keys.
 		If the database contains no data, no schema information is returned.`),
+		rawInputSchema(json.RawMessage(`{"type":"object","properties":{},"additionalProperties":false}`)),
 		mcp.WithTitleAnnotation("Get Neo4j Schema"),
 		mcp.WithReadOnlyHintAnnotation(true),
 		mcp.WithIdempotentHintAnnotation(true),
 		mcp.WithDestructiveHintAnnotation(false),
 		mcp.WithOpenWorldHintAnnotation(true),
-		mcp.WithInputSchema[EmptyInput](),
 	)
 }
--- a/internal/tools/cypher/read_cypher_spec.go
+++ b/internal/tools/cypher/read_cypher_spec.go
@@ -4,18 +4,35 @@
 package cypher
 
 import (
+	"encoding/json"
+
 	"github.com/mark3labs/mcp-go/mcp"
 )
 
 type ReadCypherInput struct {
-	Query  string `json:"query" jsonschema:"The Cypher query to execute"`
-	Params Params `json:"params,omitempty" jsonschema:"Parameters to pass to the Cypher query"`
+	Query  string `json:"query"`
+	Params Params `json:"params,omitempty"`
 }
 
 func ReadCypherSpec() mcp.Tool {
 	return mcp.NewTool("read-cypher",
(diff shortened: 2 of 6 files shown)


MacondoExpress:
Thanks for taking the time to dig into this and open a PR, the EmptyInput analysis is spot on.
On read-cypher / write-cypher: these tools are already emitting the correct schema in the current codebase. The tools/list response confirms the query is declared and required, and the params are present:
```
"inputSchema": {
  "type": "object",
  "properties": {
    "query":  { "type": "string",  "description": "The Cypher query to execute" },
    "params": { "type": "object", "description": "Parameters to pass to the Cypher query", "additionalProperties": true }
  },
  "required": ["query"],
  "additionalProperties": false
}
```
Or directly from the tests: `test/e2e/server_initialization_test.go`.

So the `WithRawInputSchema` change for those two tools isn't needed.
On `get-schema` / `list-gds-procedures`: you're absolutely right that EmptyInput produces a broken schema that requires a nonsensical "properties" argument. However, it turns out that the fix is already baked into mcp-go itself. [PR #727](https://github.com/mark3labs/mcp-go/pull/727) ("fix: always include properties field in tool JSON Schema", merged March 6, 2026) ensures toolArgumentsSchemaMarshalJSON always emits "properties": {} even for a bare NewTool with no parameters. 

We're on v0.54.0, so EmptyInput has been redundant since we bumped past v0.46.0. 

The fix on our side https://github.com/neo4j/mcp/pull/300 is simply dropping EmptyInput and the WithInputSchema[EmptyInput](), the default NewTool behaviour already produces the correct schema.

We're going to go ahead with that minimal change on our side. 
To merge a PR from an external contributor, we'd also need a signed CLA, but given the direction we're taking, this won't be necessary.
Thanks again for the detailed PR; it was useful in surfacing this.

PR #299, opened 2026-06-08
Add tools/call middleware to return error on invalid tool calls

risset:
Return an error when the tool in the tools/call request is:
- a write tool, and read-only mode is enabled (X-Neo4j-ReadOnly)
- not present in the configured tools list (X-Neo4j-MCP-Tools)

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -96,6 +96,41 @@ func NewNeo4jMCPServer(version string, cfg *config.Config, dbService database.Se
 			}
 			return filteredTools
 		}),
+		server.WithToolHandlerMiddleware(func(next server.ToolHandlerFunc) server.ToolHandlerFunc {
+			// WithToolFilter controls tool advertisement; this middleware enforces execution
+			return func(ctx context.Context, req mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+				// Guards are checked read-only first, then the configured tools list.
+				// When a request violates both, the read-only error takes precedence.
+				readOnly := mcpcontext.GetReadOnly(ctx)
+				if readOnly != nil && *readOnly {
+					mcpServer := server.ServerFromContext(ctx)
+					if mcpServer == nil {
+						// Should be unreachable: the SDK always injects the server into
+						// the context before invoking tool handler middleware.
+						return nil, fmt.Errorf("internal error: MCP server missing from context")
+					}
+
+					serverTool := mcpServer.GetTool(req.Params.Name)
+					if serverTool == nil {
+						// Should be unreachable: the SDK rejects unknown tool names before
+						// the middleware runs (see TestHTTPPerRequestToolsExecutionGuardInvalidTool).
+						return nil, fmt.Errorf("internal error: tool %q not found", req.Params.Name)
+					}
+
+					readOnlyHint := serverTool.Tool.Annotations.ReadOnlyHint
+					if readOnlyHint == nil || !*readOnlyHint {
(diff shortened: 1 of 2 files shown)


internal/server/server.go
```diff
@@ -96,6 +96,40 @@ func NewNeo4jMCPServer(version string, cfg *config.Config, dbService database.Se
...
+				}
+
+				tools := mcpcontext.GetTools(ctx)
+				if tools != nil && !slices.Contains(*tools, req.Params.Name) {
+					return mcp.NewToolResultError(fmt.Sprintf("'%s' is not in the list of configured tools", req.Params.Name)), nil
+				}
+
+				return next(ctx, req)
+			}
+		}),
```
risset, line 133: I think it would be worthwhile to refactor this and the the ToolFilter, before we consider this fully done:
- Both of them need to check if a tool is valid so they should ideally do that in the same way, like through a function that takes the context and tool name, and returns a boolean
- I find that both of them are not that readable and clear atm, due to the amount of repetitive nil-ness checks, pointer deferences etc

risset: Basically the `IsToolAllowed(ctx, req.Params.Name)` as mentioned previously

MacondoExpress (dismissed):

internal/server/server.go
```diff
@@ -96,6 +96,40 @@ func NewNeo4jMCPServer(version string, cfg *config.Config, dbService database.Se
 			}
 			return filteredTools
 		}),
+		server.WithToolHandlerMiddleware(func(next server.ToolHandlerFunc) server.ToolHandlerFunc {
```
keremgocen, line 99: We are normally leaning towards fewer comments and more self-explanatory code especially since the AI is usually a bit verbose, but in this case I think future maintainers might benefit from an explicit explanation here, such as:
`// WithToolFilter controls tool advertisement; this middleware enforces execution.`

test/e2e/http_server_per_request_tools_execution_guard_test.go
```diff
@@ -0,0 +1,174 @@
...
+			extraHeaders: map[string]string{"X-Neo4j-MCP-ReadOnly": "false"},
+			toolName:     "write-cypher",
+		},
+		{
+			name:         "When X-Neo4j-MCP-Tools is provided and contains tool being called, no error should be returned",
+			extraHeaders: map[string]string{"X-Neo4j-MCP-Tools": "read-cypher, write-cypher"},
+			toolName:     "write-cypher",
+		},
+		{
+			name:         "When X-Neo4j-MCP-ReadOnly is true, a write tool call should be blocked",
```
keremgocen, line 54: we could also benefit from a happy path read-only here:
```
extraHeaders: map[string]string{"X-Neo4j-MCP-ReadOnly": "true"},
toolName:     "read-cypher",
no errors
```
and it's a bit nit but maybe one for the precedence:
```
"X-Neo4j-MCP-ReadOnly": "true",
"X-Neo4j-MCP-Tools":    "get-schema",  // write-cypher not in list AND violates read-only
should be readOnly error, not a tool error
```
But they are both non-blocking and nice to have.

keremgocen (dismissed):
LGTM + some comments

internal/server/server.go
risset: I think, can come back to this later since it's not urgent

test/e2e/http_server_per_request_tools_execution_guard_test.go
risset: Done

keremgocen (approved):

PR #316, opened 2026-07-17
Remove minimumReleaseAge check for digest updates

keremgocen:
Eliminate the minimumReleaseAge requirement for digest and pinDigest updates in the configuration so the auto-merge can kick in once the PR is approved.

(see: https://github.com/renovatebot/renovate/discussions/39781 for the root cause, digest updates are usually missing the timestamp so our updates were left stale 14+ days until one of us merges them)

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -15,6 +15,10 @@
     "minimumReleaseAge": "3 days",
     "packageRules": [
         {
+            "matchUpdateTypes": ["digest", "pinDigest"],
+            "minimumReleaseAge": null
+        },
+        {
             "matchUpdateTypes": ["major"],
             "groupName": null
         },


renovate.json
```diff
@@ -14,6 +14,10 @@
     "separateMajorMinor": true,
     "minimumReleaseAge": "3 days",
     "packageRules": [
+        {
+            "matchUpdateTypes": ["digest", "pinDigest"],
+            "minimumReleaseAge": null
```
Liam-Doodson, line 19: From the thread you sent this looks like a nicer option no? That way it uses the timestamp if it can find it

```suggestion
            "minimumReleaseAgeBehaviour": "timestamp-optional"
```

PR #318, opened 2026-07-29
Improve HTTP MCP server observability

MacondoExpress:
Adds structured logging for HTTP-mode diagnosis: request correlation, completion lines, middleware rejections, and tool lifecycle. Sanitizes Bolt URIs via neo4j_target while keeping raw connection keys denylisted.

Highlights:

- Server-generated request_id and per-request request completed log (status, duration, auth type, headers)
- rejectRequest helper with stable reason codes on all middleware rejections
- Tool call start/complete wrapper with duration and success flag
- Initialize verification success/failure logs; guard blocks and tool-filter debug logs
- Neo4j error neo4j_error_code on database and tool error paths
- Aura db_id inferred from Bolt hostname (*.databases.neo4j.io / neo4j-dev.io)
- SafeBoltTarget in logger; log as neo4j_target (visible), not bolt_uri (redacted)
- Startup log harmonized to listen_url; url added to denylist
- Included observability fields in log (read-only, tools filter, auth type)


cmd/neo4j-mcp/neo4j-mcp
Liam-Doodson, line 1: I'm guessing this shouldn't have been committed?

internal/logger/logger.go
```diff
@@ -134,12 +134,15 @@ var sensitiveKeys = map[string]bool{
...
 
-	// Connection details
-	"uri":      true,
-	"address":  true,
-	"host":     true,
-	"port":     true,
-	"bolt_uri": true,
+	// Raw connection keys: values may contain credentials or full endpoints.
+	// For observability, use neo4j_target with SafeBoltTarget instead.
+	"uri":        true,
```
Liam-Doodson, line 139: Why do we have both `uri` and `url`? When should each one be used?

cmd/neo4j-mcp/neo4j-mcp
MacondoExpress: Oh good catch, I didn't catch I left a compiled  file there 😬

internal/logger/logger.go
MacondoExpress: Unsure, if it's used or not, similar for others it's a list of sensitive key that if found  are redacted, even before it was an initial set of sensible sensitive keys and was expanded as it. But we can also go towards the direction to keep only what is already used

Liam-Doodson: Hmm yeah I think we should probably only add if it's actually a value we're using

Liam-Doodson: If the thinking is about protecting against a future regression, maybe it's a sign we need to think about an allow list of keys that don't need redacting?

MacondoExpress: > If the thinking is about protecting against a future regression, maybe it's a sign we need to think about an allow list of keys that don't need redacting?

Fan of the idea! 
It's a bit tangential to the PR however, there is no reason for this file to be modified as part of the PR, the PR was not about the redaction mechanism, the only reason why I modified in the first place was that I've included the `safeBotTarget` as mechanism to safely print target connection w/o credential and I highlighted the impact here. But we can revert that change if that coupled the PR with redaction behaviour changes

Liam-Doodson: Hmm if it's not related then yeah maybe we just card it up for now then?

Liam-Doodson (dismissed):

MacondoExpress:
Sorry @Liam-Doodson,
My bad, but after testing one last time, I wanted to add some others logs and I've included a proper log for the the timeout case just added, plus other bits.
Can I ask a quick review 🥺 ?
The changes:
- Added request started logs with the MCP method, making initialize, tools/list, and tools/call distinguishable.
- Changed tool-start logs from INFO to DEBUG.
- Tool failures and policy blocks now use WARNING; unexpected internal errors remain ERROR.
- Removed raw Cypher queries from logs.
- Added clear timeout warnings with the affected method, phase, duration, and tool name when available.

Liam-Doodson (approved):

PR #319, opened 2026-07-30
add configurable request timeout with per-request HTTP override

MacondoExpress:
Bounds MCP requests with a timeout: 3m default, 30m maximum, set via
`NEO4J_MCP_REQUEST_TIMEOUT` or `--neo4j-request-timeout`. HTTP callers can lower it
per-request with `X-Neo4j-MCP-Request-Timeout`; larger or malformed values return 400.
Tool calls and `initialize` that run past the limit return
`request timed out after <duration>`.
Worth a close look:
- HTTP `WriteTimeout` and graceful shutdown now derive from the request timeout, so
  default shutdown goes from 65s to 3m10s.
- The deadline covers tool calls and `initialize` only, not `tools/list` or `ping`.

Apart from that the HTTP header key have a single definition space and are reused across the codebase.

### DIFF
--- a/internal/server/server_http_test.go
+++ b/internal/server/server_http_test.go
@@ -173,12 +173,6 @@ func TestHTTPServerTimeoutConstants(t *testing.T) {
 			description:   "Should be 15s to prevent slow-read attacks",
 		},
 		{
-			name:          "WriteTimeout",
-			actualValue:   serverHTTPWriteTimeout,
-			expectedValue: 60 * time.Second,
-			description:   "Should be 60s to allow complex Neo4j queries",
-		},
-		{
 			name:          "IdleTimeout",
 			actualValue:   serverHTTPIdleTimeout,
 			expectedValue: 120 * time.Second,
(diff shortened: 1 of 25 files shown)


internal/server/server_http_test.go
```diff
@@ -172,12 +172,6 @@ func TestHTTPServerTimeoutConstants(t *testing.T) {
 			expectedValue: 15 * time.Second,
 			description:   "Should be 15s to prevent slow-read attacks",
 		},
-		{
```
MacondoExpress, line 175: This test has been removed as now is covered by a dynamic timeout depending on the timeout configuration

Liam-Doodson:
Is there a corresponding PR with docs changes as this feels like something that could get missed 😬 

As an aside, have we even got a branch in the docs repo to allow us to start adding docs for v2 changes as we make them?

Liam-Doodson (approved):

MacondoExpress:
> Is there a corresponding PR with docs changes as this feels like something that could get missed 😬
> 
> As an aside, have we even got a branch in the docs repo to allow us to start adding docs for v2 changes as we make them?

We haven't create yet a branch 2.x for the documentation changes in the 2.x branch, as we have reviewed the process in the future we keep documenting at every PR, however, right now if I document this, it will be the only feature/change documented. We have have an item for the 2.x documentation, I think at this point I would prefer review the 2.x changes holistically and document them

PR #330, opened 2026-08-26
Remove deprecated environment variables and CLI flags from 2.x branch

MacondoExpress:
As for title, this PR removes the deprecated environment variables in favour of the scoped `NEO4J_MCP_*` one. It also removes the CLI args prefixes `--neo4j-*`. 
In this PR is also found the migration in place for environment variables and CLI args introduced in the 2.x and not yet released such as `--request-timeout` and `--tools`.

### DIFF
--- a/.changes/unreleased/Major-20260416-111021.yaml
+++ b/.changes/unreleased/Major-20260416-111021.yaml
@@ -2,6 +2,6 @@ kind: Major
 body: |
   HTTP mode: URL path changed from /mcp to /db/{databaseName}/mcp, which allows targeting specific databases in a multi-database Neo4j instance.
   All HTTP mode requests must now include the target database name in the URL path. The previous /mcp endpoint no longer exists.
-  NEO4J_DATABASE / --neo4j-database are rejected in HTTP mode. Setting these in HTTP mode now causes a startup error.
-  STDIO mode: NEO4J_DATABASE no longer defaults to "neo4j". This configuration option now needs to be set explicitly when using STDIO. Set NEO4J_DATABASE=neo4j explicitly to preserve the previous behaviour.
+  NEO4J_MCP_DATABASE / --neo4j-database are rejected in HTTP mode. Setting these in HTTP mode now causes a startup error.
+  STDIO mode: NEO4J_MCP_DATABASE no longer defaults to "neo4j". This configuration option now needs to be set explicitly when using STDIO. Set NEO4J_MCP_DATABASE=neo4j explicitly to preserve the previous behaviour.
 time: 2026-04-16T11:10:21.6361+01:00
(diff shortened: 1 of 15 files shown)


.changes/unreleased/Major-20260416-111021.yaml
```diff
@@ -2,6 +2,6 @@ kind: Major
 body: |
   HTTP mode: URL path changed from /mcp to /db/{databaseName}/mcp, which allows targeting specific databases in a multi-database Neo4j instance.
   All HTTP mode requests must now include the target database name in the URL path. The previous /mcp endpoint no longer exists.
-  NEO4J_DATABASE / --neo4j-database are rejected in HTTP mode. Setting these in HTTP mode now causes a startup error.
-  STDIO mode: NEO4J_DATABASE no longer defaults to "neo4j". This configuration option now needs to be set explicitly when using STDIO. Set NEO4J_DATABASE=neo4j explicitly to preserve the previous behaviour.
+  NEO4J_MCP_DATABASE / --neo4j-database are rejected in HTTP mode. Setting these in HTTP mode now causes a startup error.
```
keremgocen: should be `--database` now after this PR (I'm not sure editing previous changelogs are ok btw, I'd expect to create a new one to cover the new changes instead of editing previous ones, but maybe this is ok because it's not released yet? it kinda makes the previous commits semantically incorrect though🤷)

keremgocen: also looks like we missed .changes/unreleased/Minor-20260818-110500.yaml which announces some deprecated variables. And we said they are removed in this final changelog, so we should remove the old one since the deprecated state was never actually shipped to users, and it will create conflicting entries in the final one.

keremgocen (dismissed):
added comments to fix changelog entries, the rest is LGTM!

.changes/unreleased/Major-20260416-111021.yaml
MacondoExpress: This a good catch,  thank you! 

> also looks like we missed .changes/unreleased/Minor-20260818-110500.yaml which announces some deprecated variables. And we said they are removed in this final changelog, so we should remove the old one since the deprecated state was never actually shipped to users, and it will create conflicting entries in the final one.

That one will be automatically removed when we will release it / merge it on the main branch! These are unreleased changelog, so we're changing them before are released

keremgocen: ok, it was worth double-checking as I was not sure if a merged Minor changeset would create conflicting text when merged with Major that mentions something very similar but different, then LGTM

keremgocen (dismissed):
changes LGTM, some build is failing but not sure what the error is, worth restarting it maybe.

MacondoExpress:
> changes LGTM, some build is failing but not sure what the error is, worth restarting it maybe.

My hypothesis seems to be that we have a timeout test in a e2e test but the MacOS test in the GH CI are running slow causing the error to appear, I increased the timeout so that even the slow GH machine has time to reply!

MacondoExpress:
@keremgocen  Can I have a stamp approval (I merged a fix for the flakyness that caused the CI yesterday to fail)

